Issue #39, opened 2019-11-20
ValueError: cannot convert float NaN to integer


rajatsaxena:
I am trying to use compute_feature function using the following argument:

```
burst_kwargs = {'amplitude_fraction_threshold': .2,
                'amplitude_consistency_threshold': .5,
                'period_consistency_threshold': .5,
                'monotonicity_threshold': .8,
                'N_cycles_min': 3}

narrowband_kwargs = {'N_seconds': .5}
```

but I keep on running in the following error. Any suggestions on how to fix it?

```
Traceback (most recent call last):

  File "<ipython-input-61-da442dc80175>", line 6, in <module>
    hilbert_increase_N=True)

  File "C:\ProgramData\Anaconda3\lib\site-packages\bycycle\features.py", line 128, in compute_features
    zeroxR, zeroxD = find_zerox(x, Ps, Ts)

  File "C:\ProgramData\Anaconda3\lib\site-packages\bycycle\cyclepoints.py", line 202, in find_zerox
    zeroxD[i] = Ps[i] + int(np.median(_fzerofall(x_temp)))

ValueError: cannot convert float NaN to integer
```

voytek:
Can you send us a snippet of the data that you're analyzing when you run into this error?

@nschawor have you  seen this before?

srcole:
Hey @rajatsaxena! yeah, it's hard to tell from there. Like Brad said, if you attach a stand-alone script that reproduces the error, that'll help us figure it out. If it includes code to simulate some data that reproduces the issue, that'll be nice. Otherwise, a link to some sort of text file would work too.

rajatsaxena:
Here is the link to the data: https://drive.google.com/file/d/1Fdky_-ARXetMIVa59jA6KqpaWZ-0g_nq/view?usp=sharing along with the code snippet to recreate the error. 


```
import numpy as np
from bycycle.features import compute_features

signal = np.load('ob.npy')
Fs = 3000
f_gamma = (40,100)

burst_kwargs = {'amplitude_fraction_threshold': .2,
                'amplitude_consistency_threshold': .5,
                'period_consistency_threshold': .5,
                'monotonicity_threshold': .8,
                'N_cycles_min': 3}
narrowband_kwargs = {'N_seconds': .5}

df = compute_features(signal[:2500000], Fs, f_gamma,
                      center_extrema='T',
                      burst_detection_method='cycles',
                      burst_detection_kwargs=burst_kwargs,
[...]

PR #40, opened 2019-12-08
Assign dummy zerocrossing to avoid error

srcole:
Issue #39


TomDonoghue:
this seems to fix the bug reported in #39, and seems sensible to me, so I'm going to merge it in

Issue #25, opened 2018-11-29
More thorough testing


srcole:
Code coverage is currently at 63%, which @TomDonoghue says is "not sufficient."

More tests would help assure that all functions behave as expected in various situations.

TomDonoghue:
Closed with #50

PR #50, opened 2020-04-30
[ENH] Increased coverage from 63 to 95%

ryanhammonds:
Addressed #25

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,22 @@
-dist/
-bycycle.egg-info/
-build/
+# Ignore notebook checkpoints
 .ipynb_checkpoints/
-*.pyc
+
+# Ignore cache files
+__pycache__/
+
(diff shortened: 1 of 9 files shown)


TomDonoghue:
Okay, @ryanhammonds - so I had a look through this, and directly edited for some aesthetic tweaks, and also to add the `summary` file, like we have in other modules, and update `gitignore`. 

In terms of the tests: to be honest, these pytest constructs are really cool, but not something I am super familiar - so basically I read through as a way to learn a bit more about them. I did the checks through the files that nothing looks weird to me, and I can run the test locally, and travis is happy, all while greatly extending our covereage - so this all looks good to me! Merging in!

ryanhammonds:
Cool, @TomDonoghue, glad this got merged! I used the parametrize decorator to help cover as many lines as possible. There may have been a simpler way.

Issue #32, opened 2019-03-13
1.0 Refactor & Updates


TomDonoghue:
Note: this issue has been updated to re-organized into an overall plan for refactors and updates for an API breaking, major update 1.0.0 release. 

This refactor is modeled on the refactor for NDSP for the 2.0 release. The main PR for that was in https://github.com/neurodsp-tools/neurodsp/pull/150. Checking some of the commits / diffs / updates, if they are readable, should hopefully be something of a model for this update. 

To have an easier organizations of the updates, and review process, let's do a one-thing-at-a-time process, so probably each of the following points should be 1 PR (or possibly more). I think it makes sense to do these in order, top to bottom.

#### Scoping & Dependency Structure (#31)
- [x] remove the sim code, and associated test code, etc (this now lives in NDSP)
- [x] remove the current filter code, and add NDSP as a dependency for this stuff, updating to use the now updated versions of the same code, from NDSP
- [x] work through the burst file, and remove copied code. Burst procedures are to be considered  a core functionality of ByCycle, but for the moment we need to balance shared code between NDSP / ByCycle. 

#### Code Style
- [x] consolidate var names to be consistent with how NDSP has things, and also to match code style (for example, snake_case for all variables, all vars at least 2 characters, etc)
    - for example `x` -> `sig` and `Fs` -> `fs`
    - run a linter, and update code accordingly
- [x] check through for code & doc style (for example, capitalization & punctuation in docstrings)

#### Modularize & Code Updates
- [x] move any plotting functions to their own /plts/ folder
    - Plot code that is written out to create plots in the tutorials, could also be made into functions that are added to the module. 
- [x] split out some of the larger functions, increasing the modularity of the code, and providing more access to some meta-data components (this could be done file-by-file. Example: #46)
    - for example, separating out the metadata for peak, trough and zero-crossings from the analysis, so that it is independently accessible and useable (comment from Erik)
    - this will also include updating and splitting up the tests
- [x] For the new modularized organization, add doctest examples

#### Tutorials / Documentation
- [x] Update tutorials, reflecting all the new code updates. Work through and see if we should extend, add new things to the tutorials, make sure we have good coverage of ideas, etc.
- [x] Update the README. The code overview can probably be minimized, and point more directly to the documentation website.
- [x] Add glossary (#62) & reference pages to the docsite
- [x] Add helper function to download data for the tutorials / examples (so they can be downloaded & run)
[...]

PR #54, opened 2020-05-20
[CLN] Scoping & Dependency Structure

ryanhammonds:
This is related to the first two points in #32. `bycycle.sim` and `bycycle.filt` have been replaced with `neurodsp` dependencies in the module, tests, and notebooks. Intershpinx was also configured to link to the neurodsp site. [Here](https://github.com/ryanhammonds/bycycle/blob/explores/filter_sig.ipynb) is a notebook that shows the neurodsp funcs/replacements are equivalent.

Replacing `bycycle.burst` will be added to a separate PR since doing so was more complicated/messy.


TomDonoghue:
I've had a look at this one, and everything looks good to me. 

Since stylistic nitpicks are not an issue here (things for later), I checked by:
- checking a local install, running tests, and building the docsite
- reading the code the check for any quirks
- comparing the example outputs in the Tutorials to those from 0.1.3.

Everything looks good to me, and looks totally equivalent when calling these functions from NDSP. 

Approved from me - ready to merge!

Issue #31, opened 2019-03-12
Move time-series actions to NDSP & add dependency?


TomDonoghue:
We've been talking a bit about making a consistent 'shared code world' across VoytekLab tools, as well as clean scoping of what each module does. 

As a heuristic, this is the proposed outline:
- FOOOF: operates on frequency representations
- NDSP: operates on time series
- ByCycle: operates on cycles

As is, ByCycle sorta of breaks this, by including a bunch of time-series operations, and this includes a lot of copied code, that is messy to maintain, and inconsistencies in scope (example - the best burst approach lives here, even though it's not, by itself, a cycle operation, and meaning that neurodsp/burst isn't the best place to do bursting from). 

So, the proposal, if we want to follow this, is that general time-series function should be consolidated into NDSP, and ByCycle shouldn't have it's own copies, but rather have NDSP as a dependency, and use the same general functions from there. 

I think this has benefits in terms of maintainability and clear scoping of projects, and I don't think there's any particular issue with having ByCycle depend on NDSP. 

This would mean:
- bycycle/filt: would be removed, and the same functions would be used from NDSP
- bycycle/burst: the burst detection procedure here would be moved to NDSP, and then used from here. 
- bycycle/sim: would be removed, and simulating time series (including with cycle-by-cycle variation) would be considered a time-series operation, that lives in NDSP.

cc: @srcole for thoughts on this.

srcole:
yup, that makes sense.

I think the original reason for having the copied code was that neurodsp was going through a lot of changes as I was developing bycycle ;)

now that they're more stable, yeah, it would totally be better to reduce duplicate code like you outline

TomDonoghue:
Done in #54

PR #55, opened 2020-05-22
[WIP] Scoping of burst detection module

### LINKED ISSUE #46: Modularize  bycycle.burst.detect_bursts_cycles
It would be desirable to compute the ```df.is_burst``` column without having to supply the data for easier hyperparameter tuning as well as to recompute this column when dealing with epoched data. 

This computation is performed within the ```burst.detect_bursts_cycles``` function, which takes the burst features dataframe as well as the data vector ```x```.  The data is needed to compute monotonicity, but it could be done in ```features.compute_features``` or further modularized somewhere else, so ```burst.detect_bursts_cycles``` does not require ```x``` as a parameter anymore. This would simplifying rerunning this function, without having to load the data along with the dataframe.

ryanhammonds:
### Background
This is related to the first two points in the larger refactor, #32, and the point brought up in #31, which suggested removing `bycycle.burst` in exchange for equivalent functionality from neurodsp.

This PR should also include changes to help resolve #46. This issue suggested removing the `x` argument, or signal, from `burst.detect_bursts_cycles` . This argument is currently used to calculate monotonicity, but can be moved to `compute_features` or elsewhere. 

### Original Changes
This PR has so far removed dual amplitude threshold burst detection, and used an identical neurodsp implementation. This change can be compared in a [neurodsp dependent notebook](https://github.com/ryanhammonds/bycycle/blob/explores/burst_refactor.ipynb) and an [original notebook](https://github.com/ryanhammonds/bycycle/blob/explores/burst_original.ipynb). The "neurodsp dependent notebook" was ran with a  bycycle installation from this PR whereas the "original notebook" was ran with a master installation. This was to compare the effect of this PR. These notebooks show that the changes implemented so far do not change the burst detection or dataframe output.

### Other Suggested Changes

- @TomDonoghue has suggested keeping `bycycle.burst.detect_bursts_cycles` in bycycle, rather than moving to neurodsp.
- @nschawor has suggest modularizing `bycycle.burst.detect_bursts_cycles` (#46) by removing the signal argument. I will go ahead and work on implementing this. 
- Move `bycycle.burst.plot_burst_detect_params` to a separate plotting module (a point in #32).

These changes would leave `detect_bursts_cycles` and `detect_bursts_df_amp` (which is now limited to moving results from `neurodsp.burst.detect_bursts_dual_threshold` to a pandas df) functions within `bycycle.burst`.

I'm open to any comments or suggestions about the scope or modularity of `bycycle.burst`   :slightly_smiling_face:


TomDonoghue:
In terms of removing copied code, and in this case, organizing that the dual-amp approach is a time-series operation to be housed in NDPS, and cycle-burst lives in ByCycle, this looks good to me. 

To have small, modular PRs (that are easier to review, and leave a better paper trail):
- this PR should do the scoping, dropping the stuff that's in NDSP
- refactors, related to #46 is a different task, which can go into a different PR (and you can move the comments above about refactoring into a new PR). 

So, reviewed in terms of burst scope, can you update the test files to pass with the new organization of burst? After that, this should basically be good to merge, unless there are other major comments on the organization so far.

nschawor:
Brief comment because of tag. In my view, removing ```twothresh_amp``` from bycycle is not useful. 

1. In the future, more work is needed on **hyperparameter tuning** and for this direct comparison to more classical approaches are useful, probably requiring modifications of ```twothresh_amp```, easier to do if this is within one repository.
2. This separation between operations timeseries and cycles sounds neat in theory, but not really aiding the **functional purpose** of this package (there will always be overlap because input is timeseries and many interesting procedures to extract bursts will operate on timeseries because of long-range temporal correlations).  It may be of interest to add more approaches for burst detection (like the hidden markov model ones) to bycycle to improve investigation of waveform features. Having a **comprehensive package** encompassing many different approaches for methods comparison for this would be very nice, not scattered across different code bases.



If one is concerned about code duplication it makes more sense to delete the dual threshold function from neurodsp.

TomDonoghue:
Yeh, organizing burst code I think is the trickiest part of scoping between modules, since bursting is like a 'first-class operation' in ByCycle (arguably a cycle-operation regardless of approach). 

In this case, 2amp is already in NDSP, and the particular implementation in ByCycle needs either replacing or dropping (it's old / out of sync with current API, etc). Right now, since 2amp isn't planned for current changes, replacing it would just be copied code. 

Relatedly, when we move on to refactors, I think we need to have more of a plug-in approach (be able to apply various burst-detection approaches). For example the `burst_detection_method` input in `compute_features` should accept a callable. This will make it easier to try and apply different and custom burst detection, regardless of where they live, and try new approaches, and merge in working procedures to the stable release. 

So, my proposal for now, in terms of module development:
- complete this PR, as we need to change this code anyways
- for the moment (while implementations are identical) call 2amp from NDSP
- through the refactors, make sure burst approaches are more modular / plug-in style
- in future development work bursts in ByCycle, if the 2amp approach needs using / investigating / tweaking, it can be dropped back in to dev, explored, and then if a useful difference from standard is found, it can be merged into master (at this point, it wouldn't be a duplicate).

tl-dr: yeh, mostly agree. it still needs updating, let's drop it for now, make burst detection approaches more general, and it can be dropped back in if it does change.

nschawor:
- Since this is a structural PR, it would be best to figure this out in a good way at this point instead of quickly closing. IMHO the result of this PR makes the organization of bycycle definitely worse.
- I would keep dual threshold in bycycle since core burst detection method.
- If I understood correctly, PR removes the possibility to return the amplitudes, which is nice to have (otherwise one has to rerun ```amp_by_time``` which requires filtering again -> inefficient)

TomDonoghue:
>     * Since this is a structural PR, it would be best to figure this out in a good way at this point instead of quickly closing. IMHO the result of this PR makes the organization of bycycle definitely worse.

I agree on figuring this out, but I did propose a plan, and I didn't say "lets just close it"....? I don't understand why this plan makes anything worse. 

>     * I would keep dual threshold in bycycle since core burst detection method.

I don't think it really matters where the current implementation is used from though? I agree on a plan to develop bursting in ByCycle. If 2amp changes it will live here. Developing a new stable release is different from the future development plans. Keeping the current code is not viable, and having literal copied code from NDSP doesn't make sense to me.

>     * If I understood correctly, PR removes the possibility to return the amplitudes, which is nice to have (otherwise one has to rerun `amp_by_time` which requires filtering again -> inefficient)

I'm unclear on this point. I agree if this changes behaviour that's important, but this literally just changes where the function gets called from, and doesn't change anything about what is available, I think (?). In what way does this change accessibility to the amplitudes?

ryanhammonds:
I'll save refactors for a different PR. After reading through this thread, it seems like the main decision involves where `twothresh_amp` should live. If it is bycycle, as suggested by @nschawor, the changes in this PR can be ignored and the PR closed without merging. If it is ndsp, as suggested by @TomDonoghue, then the PR can be merged. If it lives in both bycycle and ndsp, it would be useful to ensure that the two functions are synchronized and function identically. This may become difficult and it may be useful to drop from one package or the other.

As for this:
> > ```
> > * If I understood correctly, PR removes the possibility to return the amplitudes, which is nice to have (otherwise one has to rerun `amp_by_time` which requires filtering again -> inefficient)
> > ```
> 
> I'm unclear on this point. I agree if this changes behaviour that's important, but this literally just changes where the function gets called from, and doesn't change anything about what is available, I think (?). In what way does this change accessibility to the amplitudes?

`twothresh_amp`  allows amplitudes to be returned using a `return_amplitude` argument (defaults to false). This isn't implemented in ndsp's `detect_bursts_dual_threshold`, but would be easy to. Returning amplitudes isn't required for `compute_features`  -> `detect_bursts_df_amp`, but may be useful outside of this context.

TomDonoghue:
The version in NDSP is already refactored / cleaner, so I don't think it makes sense to drop this PR entirely, because then we would have to redo a refactor that is already done, and end up with two copies. For compatibility, we can't wholesale drop 2amp from NDSP right now, but we could mark it for deprecation and plan to move later. For the purpose of a stable release, I think merging this PR makes the most sense, because we do not want two identical copies, and this still doesn't limit anything about development or future burst plans. 

I'm a little unclear if returning amplitudes matters. What's the use case? I haven't been able to find a use of the returning amplitudes from `twothresh_amp` in any of the project repos.

nschawor:
> After reading through this thread, it seems like the main decision involves where `twothresh_amp` should live. If it is bycycle, as suggested by @nschawor, the changes in this PR can be ignored and the PR closed without merging. If it is ndsp, as suggested by @TomDonoghue, then the PR can be merged. If it lives in both bycycle and ndsp, it would be useful to ensure that the two functions are synchronized and function identically. This may become difficult (at least to continually ensure this) and it may be useful to drop from one package or the other.

thank you for the summary. yes, a consequence of all burst related things residing in bycycle would be that the counterparts in neurodsp should be removed. the main advantage is conceptual cohesiveness and clarity regarding the functional purpose of this package.

TomDonoghue:
@nschawor are you recommending we drop burst from NDSP right now, and potentially break people's code? Because that is going to be a no from me. Stability of module development is important.

nschawor:
hi, I don't have any opinion on neurodsp time line. freeze and start deprecating process?

so: replace the version in bycycle with the newest version, looking at the time scale of changes to this function, a short-time duplication of code is pretty much a non-issue.

TomDonoghue:
Okay, we are agreed on burst living in ByCycle, and we can set out a process for deprecating things in modules. We still want to get rid of this old implementation, so I'm going to merge this is in. We can add always add it back after, as we work through 1.0. Using an import is a non issue, and doing so temporarily means we won't fall out of sync as we do refactors.

PR #57, opened 2020-06-04
[MRG] Plotting refactors.

ryanhammonds:
This PR refactor plotting functions to their own module, `bycycle.plts`, a plan outlined in #32 . Two new plotting functions/tests were added: `plot_cyclepoints` and `plot_cycle_features`. Sphinx/examples/tutorials were all updated for these new functions. I also made `plot_burst_detect_params` a tiny bit more modular, although it still functions identically as in 0.1.3, with the exception of being imported from `bycycle.plts`.

This PR is dependent on a neurodsp PR that I'll need to merge first.


ryanhammonds:
Hi @voytek, @nschawor, @sydney-smith, I spoke with Tom today about a plotting change for burst detection parameters. Currently, burst detection is plotted from peak to peak using interpolation. This is shown in the second plot. However,  burst detection params are constant for any cycle. Does it make sense to plot these parameters from trough to trough (or peak to peak if trough-centered), so the burst param is constant over the course of a cycle, as in the third plot?

Screenshot_20200609_143509

TomDonoghue:
Just a note that we largely wanted to sanity check this change, and make sure we weren't missing something about the old plot approach - so if anyone has any comments or suggestions, feel free to say more - but if this looks good / innocuous, then feel free to just drop a thumbs up.

sydney-smith:
I actually really like this idea. In a recent project, it seemed like
"first" real cycle of the burst was sometimes getting skipped because of
this. And this was especially an issue because this first cycle was the
focus of the analysis.

voytek:
Chiming in late, but this is a plotting change only, right? So it doesn't address the issue that @sydney-smith mentions?

ryanhammonds:
Hi @voytek, yes, this is just a plotting change. I will make a separate issue for the problem Sydney mentioned. I looked into a bit yesterday, and the reason the first cycle is skipped is due to nan values for amplitude and period consistency for the first cycle. I need to dig into it a bit more to find a fix.

For the plotting change, Natalie preferred interpolation and mentioned that cycle features aren't really constant over a cycle but are computed as so. I left the interpolated lines, as in the second graph, but made sure the green coloring covers full cycles.

TomDonoghue:
Hmm - interpolated lines and shaded cycles sounds a bit weird to me, because then the shading doesn't track the threshold directly, and that seems an extra level to interpret, and mixed messages between plotting by cycle or by timepoints. 

I agree that the values aren't constant - but what we're plotting are our measures (not the underlying truth, which we don't have access to), and when Ryan and I talked about interpolation we thought that we can only plot what we measure, and can note that this is a low-resolution measure, but we have no reason to expect that the intra-cycle variation is anything like the linear interpolations we currently have (so that could be misleading). So, overall, cycle level plotting is our literal measure, without assumptions of intra-cycle levels that we do not currently have access to. 

@nschawor - what's the reason to keep interpolation?

nschawor:
The main benefit I see is the highlighting that the cycle features are not constant over one cycle (not linear, but thinking of it as one approximation term improvement against only the constant one), which may protect against misconceptions like present in the neurodsp cycle based burst simulation. 

Possibly, a good standard labeling of figure components could be helpful here & also highlighting regions in the top time series graph.

TomDonoghue:
Since the purpose of these plots is to show the estimates of cycle properties, and when they drop below defined threshold, both of which are defined on a cycle by cycle basis (one value / decision per cycle) I still vote for the non-interpolated version as a reflection of the measurements here. We should of course note in tutorials / documentation that there can be intra-cycle variation - but we don't know what it looks like and I don't think we want to imply it's continuous / linear variation. 

More labeling can definitely be added to these plots. 

@nschawor - I don't know what you are referring to for NDSP bursts....? If there is some kind of problem with that, you should open an issue on NDSP.

ryanhammonds:
What if I made an ‘interp’ argument that way both plotting options are available?

Test are passing now for this PR and it should almost be ready to review/merge. I just need to update the tutorials and api.

bycycle/plts/features.py
```diff
@@ -0,0 +1,73 @@
...
+
+    if xlim:
+        ax.set_xlim(xlim)
+
+    feature = df if not isinstance(df, pd.core.frame.DataFrame) else df[burst_param]
+
+    ax.hist(feature, bins=bins, color=color, alpha=.5)
+
+
+
```
TomDonoghue, line 69: tooooo many blank lines :)

bycycle/plts/features.py
```diff
@@ -0,0 +1,73 @@
...
+import numpy as np
+import pandas as pd
+import matplotlib.pyplot as plt
+
+###################################################################################################
+###################################################################################################
+
+
+def plot_feature_hist(df, burst_param, ax=None, bins='auto', **kwargs):
+    """ Plot a histogram for a cycle feature.
```
TomDonoghue, line 12: extra leading space.

bycycle/plts/features.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Parameters
+    ----------
+    df : pandas.DataFrame or 1d array
+        Dataframe output from :func:`~.compute_features` or a 1d array.
+    burst_param : str
+        Column name of the parameter of interest in ``df``.
+    ax : matplotlib.Axes, optional, default: None
```
TomDonoghue, line 20: I have tended to consistently put `ax` as the last argument (except for kwargs), and I think we might as well keep this up (unless you disagree). If it sounds good, check and update across all plot funcs.

bycycle/plts/features.py
```diff
@@ -0,0 +1,73 @@
...
+
+    - ``xlabel``: str, default: ``burst_param``
+    - ``figsize``: tuple of (float, float), default: (10, 10)
+    - ``color``: str, default: 'k'
+    - ``xlim``: tuple of (float, float), default: None
+    - ``fontsize``: float, default: 15
+
+    Examples
+    --------
+    See the `feature distribution example <http://bycycle-tools.github.io/bycycle/auto_examples/plot_theta_feature_distributions.html#plot-feature-distributions>`_.
```
TomDonoghue, line 40: I don't think we need this, really. In the API listing we list out examples using this function. Otherwise, it seems odd to have a link here, but not for other functions.

bycycle/plts/burst.py
```diff
@@ -0,0 +1,227 @@
...
+    tlims : tuple of (float, float), optional, default: None
+        Start and stop times for plot.
+    figsize : tuple of (float, float), optional, default: (15, 3)
+        Size of each plot.
+    plot_only_result : bool, optional, default: False
+        Plot only the signal and bursts, excluding burst parameter plots.
+    burst_params : list of str, optional, default: None
+        The names of the ``df`` burst detection parameter columns. These names may differ from the
+        keys in ``osc_kwargs``. This is only needed for non-default burst parameters.
+    interp : bool
```
TomDonoghue, line 44: optional, default: True

Also, I think it could be a little more descriptive, like:
"If True, interpolates between given values. Otherwise, plots in a step-wise fashion."

bycycle/plts/burst.py
```diff
@@ -0,0 +1,227 @@
...
+
+    Parameters
+    ----------
+    df : pandas.DataFrame
+        Dataframe output of :func:`~.compute_features`.
+    sig : 1d array
+        Time series to plot.
+    fs : float
+        Sampling rate, in Hz.
+    osc_kwargs : dict
```
TomDonoghue, line 32: I'm not so sure about the name of this inputs (and, relatedly, the `burst_params` input below). We don't need to maintain the same labels from before (this is a breaking update anyways). 

If this is the same as `burst_detection_kwargs`, I think it should be named the same. And `burst_params` seems like it's custom names or something, so I think there would be a better name there. Maybe just `burst_param_labels`?

bycycle/plts/burst.py
```diff
@@ -0,0 +1,227 @@
...
+
+            if cyc[column] < osc_kwargs[osc_key]:
+                axes[0].axvspan(times[cyc['sample_last_' + side_e]],
+                                times[cyc['sample_next_' + side_e]],
+                                alpha=0.5, color=color, lw=0)
+
+        # Plot each burst param on separate axes
+        if not plot_only_result:
+
+            ylabel = osc_key.replace('_threshold', '').replace('_', ' ').capitalize()
```
TomDonoghue, line 132: I like this for a generalized updating to pretty labels!

bycycle/plts/burst.py
```diff
@@ -0,0 +1,227 @@
...
+    sig = zscore(sig)
+
+    # Determine time array and limits
+    times = np.arange(0, len(sig) / fs, 1 / fs)
+    tlims = (times[0], times[-1]) if tlims is None else tlims
+
+    # Determine if peak of troughs are the sides of an oscillation
+    _, side_e = get_extrema(df)
+
+    df_lim, sig_lim, times_lim = apply_tlims(df, sig, times, fs, tlims)
```
TomDonoghue, line 80: I'm not totally clear on why you maintain two copies of things (like `df` and `df_lim`). It seems like if and once things are trimmed, this can be used across the board, right? 

For example below, I'm not sure why 
`plot_cyclepoints(df, sig, fs, ax=axes[0], tlims=tlims, plot_zerox=False,...)`
can't be:
`plot_cyclepoints(df_lim, sig_lim, fs, ax=axes[0], tlims=None, plot_zerox=False,...)`

And so on. If there is a reason this doesn't work, we might want to add a comment and note in this function as to the reasons and so on, and/or think about updating wtv would be needed such that this could work.

bycycle/plts/burst.py
```diff
@@ -0,0 +1,227 @@
...
+    df, sig, times = apply_tlims(df, sig, times, fs, tlims)
+
+    # Plot burst param
+    if interp:
+
+        plot_time_series([times[df['sample_' + center_e]], tlims], [df[burst_param], [thresh]*2],
+                         ax=ax, colors=['k.-', 'k--'], xlim=tlims, xlabel=xlabel,
+                         ylabel="{0:s}\nthreshold={1:.2f}".format(ylabel, thresh), **kwargs)
+
+    else:
```
TomDonoghue, line 204: I think perhaps a little more incode comments here could be useful to briefly describe what is happening.

bycycle/tests/test_plts_cyclepoints.py
```diff
@@ -0,0 +1,16 @@
...
+from bycycle.plts.cyclepoints import plot_cyclepoints
+from bycycle.tests.utils import plot_test
+
+###################################################################################################
+###################################################################################################
+
+@plot_test
+@pytest.mark.parametrize("tlims", [None, (0, 1)])
+def test_plot_cyclepoints(tlims, sim_args):
+    """Test plotting extrema/zero-crossings."""
```
TomDonoghue, line 14: For consistently, we generally don't add docstrings for test funcs, so since this one doesn't say anything special, I think we could just drop it.

bycycle/tests/test_plts_utils.py
```diff
@@ -0,0 +1,47 @@
...
+    df = sim_args['df']
+    sig = sim_args['sig']
+    fs = sim_args['fs']
+
+    times = np.arange(0, len(sig) / fs, 1 / fs)
+    tlims = (1, 2)
+
+    df_short, sig_short, times_short = apply_tlims(df, sig, times, fs, tlims)
+
+    print(df_short['sample_next_trough'].values)
```
TomDonoghue, line 22: I presume these are a holdover from exploring the testing - but presumably we don't need the print lines?

bycycle/plts/features.py
```diff
@@ -0,0 +1,73 @@
...
+###################################################################################################
+###################################################################################################
+
+
+def plot_feature_hist(df, burst_param, ax=None, bins='auto', **kwargs):
+    """ Plot a histogram for a cycle feature.
+
+    Parameters
+    ----------
+    df : pandas.DataFrame or 1d array
```
TomDonoghue, line 16: If this can be mixed type, then the name `df` seems odd. What about just calling it `data` or something?

bycycle/plts/features.py
```diff
@@ -0,0 +1,73 @@
...
+    # Axis style
+    ax.set_xlabel(xlabel, size=fontsize)
+    ax.set_ylabel('# cycles', size=fontsize)
+
+    if xlim:
+        ax.set_xlim(xlim)
+
+    feature = df if not isinstance(df, pd.core.frame.DataFrame) else df[burst_param]
+
+    ax.hist(feature, bins=bins, color=color, alpha=.5)
```
TomDonoghue, line 66: `alpha` is probably something we can catch as a kwarg input, so it can be more easily tuned.

bycycle/plts/cyclepoints.py
```diff
@@ -0,0 +1,99 @@
...
+    -----
+    Default keyword arguments include:
+
+    - ``figsize``: tuple of (float, float), default: (15, 3)
+    - ``xlabel``: str, default: 'Time (s)'
+    - ``ylabel``: str, default: 'Voltage (uV)
+
+    """
+
+    rcParams['lines.markersize'] = 12
```
TomDonoghue, line 51: Won't this have side-effects? Like, effecting all subsequent plots? I'm not sure we want to be updating rcParams in here (though I am also a little naive to it's scope / how it works - so maybe it doesn't matter).

bycycle/plts/cyclepoints.py
```diff
@@ -0,0 +1,99 @@
...
+    center_e, side_e = get_extrema(df)
+
+    df, sig, times = apply_tlims(df, sig, times, fs, tlims)
+
+    # Extend plotting based on given arguments
+    x_values = []
+    y_values = []
+    colors = []
+
+    if plot_sig:
```
TomDonoghue, line 75: This is a cool & efficient way to organize the data for plotting, but see the comment elsewhere about adding an array function.

doc/api.rst
```diff
@@ -55,14 +55,17 @@ Burst Detection Functions
     :toctree: generated/
 
     detect_bursts_cycles
-    detect_bursts_df_amp
 
 Plotting Function
 =================
 
-.. currentmodule:: bycycle.burst
+.. currentmodule:: bycycle.plts
```
TomDonoghue, line 62: Can you update the heading above to "Plotting Functions"

bycycle/plts/burst.py
```diff
@@ -0,0 +1,227 @@
...
+    - If plot_only_result = True: return a plot of the burst detection in which periods with bursts
+      are denoted in red.
+
+    - If plot_only_result = False: return a list of the fig handle followed by the 5 axes.
+
+    - In the top plot, the raw signal is plotted in black, and the red line indicates periods
+      defined as oscillatory bursts. The highlighted regions indicate when each burst requirement
+      was violated, color-coded consistently with the plots below.
+
+     - blue: amplitude_fraction_threshold
```
TomDonoghue, line 60: This has a rendering issue with sphinx. I think it's missing an indent.

TomDonoghue:
Okay, thanks @ryanhammonds for tons of hard work on this one!

I did a review in which I read through all the code, and checked the rendered versions of the site & tutorials, and left a pile of comments accordingly. As always, feel free to ask about anything, or down vote anything you want to recommend against - and we can also chat on things.

I didn't really sit with these functions and play with them with open ended data, and stress test them in that way. So my proposal would be if you want to check and update from the current comments, after that, tag me back in, and I can scan the changes and at that point I'll try the plots out a bit more. 

A few notes:
- If you can also run a linter, mine pulled up a few nitpicks, and I didn't note them all directly. 
- The update for installing the dev version of NDSP is cool (good idea to get tests running here!), but I do think we should be able to release NDSP before we merge this, to avoid having a weird install approach on master. 
- One thing worth a check through at the end is that we have a consistent API (common params always passed with the same names and in the same order), and common docstring conventions (the same kind of functions described in the same way). I think we're pretty good on this, but something we can sanity check on the next round as well. 
- We do here have a different approach to applying plot style & defaults than in NDSP (& FOOOF). That's not an issue per se, but since we already depend on NDSP, we could import and use `plot_style`, as done there. Not sure if this is a good idea or not tbh, but it's an option.

tutorials/plot_1_bycycle_philosophy.py
```diff
@@ -100,32 +94,29 @@
...
-plt.xlim(tlim)
-plt.tight_layout()
-plt.show()
 
-####################################################################################################
+color_cycle = plt.rcParams['axes.prop_cycle'].by_key()['color']
+
+fig, axes = plt.subplots(figsize=(15, 6), nrows=2)
+
+plot_time_series(times, amps, ax=axes[0], colors=color_cycle, xlim=(2, 6))
```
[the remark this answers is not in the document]
ryanhammonds: `plot_time_series` accepts a `xlim` but not `tlim`. Maybe the ndsp funcs could accept both `xlim` or `tlim`, interchangeably. Or I can change `tlim` in bycycle to `xlim`. I can make the change either way you prefer.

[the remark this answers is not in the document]
TomDonoghue: Ohh, good point! 
What I was thinking of when I wrote this comment is just to write `xlim=tlim` in these lines, since we have the `tlim` variable declared. But now that you bring it up, ByCycle does use a `tlim` label for what is basically a `xlim` thing right? 

My vote would be consolidate on using `xlim` for plotting. Where-ever possible, I think we want to be consistent with the matplotlib API. 

The quirk is ByCycle is that we also have the functions that update DF values based on time ranges. For those functions, `xlim` isn't a good name. Maybe for visualizations, call it `xlim`, and for DF functions, call it `time_range` or similar?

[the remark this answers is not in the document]
ryanhammonds: I'll switch to xlim in the plotting funcs.

Bycylce actually doesn't use tlim at all outside of plotting, which I didn't realize until now. All other functions expect signal/times to be limited prior to function calls.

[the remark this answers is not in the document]
TomDonoghue: Okay, perfect, that all sounds good then!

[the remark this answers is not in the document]
TomDonoghue: I just double checked - `plots/utils/apply_tlims` uses tlims - but it seems to me that this is a general utility function that could be used outside plotting. 

What do you think of updating to something like:
`def restrict_df_time_range(df, fs, time_range):`
And putting it into a general utility function?

Related, I sorta think that function can operate on DF's without also updating signals. A separate utility, if needed, could do `restrict_sig_time_range`. (Names are quick guesses - feel free to update).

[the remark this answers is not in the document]
ryanhammonds: Good idea! I think moving it to a general utility makes since now that I've noticed there is not a `xlim` arg in `compute_features`, etc. I also think two separate funcs, one for `df` and a second for `sig`/`times`  would be helpful for time limiting in `compute_features`.

ryanhammonds:
Hi @TomDonoghue, I have push a commit (that should have really been several commits) addressing all of your comments in your review. Here is a summary of the changes, or what the commit log should've looked like:

- `plot_instantaneous_measure` used in tutorials
- `tlim` -> `xlim`
- `apply_tlims` split into `limit_df` and `limit_sig_times`, and moved to general utils
- `plot_cyclepoints` split into `plot_cyclepoints_df` and `plot_cyclepoints_array`
- `plot_features_catplot` added, it only uses mpl but is same as seaborn's catplot.
- `ax` moved to last argument.
- `burst_detect_kwargs` are now the same as the df column names, but with an appended '_threshold'

  - `burst_params` removed as arg from `plot_burst_detect_summary`

- Test updated to save plots to 'test_files`
- Other changes related to typos/linting/docs/sphinx/comments/

bycycle/utils.py
```diff
@@ -0,0 +1,90 @@
...
+def limit_df(df, fs, xlim):
+    """Restrict dataframe to be within time limits.
+
+    Parameters
+    ----------
+    df : pandas DataFrame
+        Dataframe output of :func:`~.compute_features`.
+    fs : float
+        Sampling rate, in Hz.
+    xlim : tuple of (float, float), optional, default: None
```
TomDonoghue, line 18: `optional, default: None` can be removed

bycycle/utils.py
```diff
@@ -0,0 +1,90 @@
...
+
+    return df
+
+
+def limit_sig_times(sig, times, xlim):
+    """Restrict signal and times to be within time limits.
+
+    Parameters
+    ----------
+    sig : 1d array
```
TomDonoghue, line 48: what do we think about ordering for `sig` and `times`. In plot funcs, I think we generally have `times` first, then `sig`. It might make sense for this to follow that, and so reverse order?

bycycle/utils.py
```diff
@@ -0,0 +1,90 @@
+"""Plotting utility functions."""
+
+import numpy as np
+
+###################################################################################################
+###################################################################################################
+
+
+def limit_df(df, fs, xlim):
```
TomDonoghue, line 9: `xlim`, as a name, is really related to plotting, and these functions are now organized as general utilities.
Also, the current API makes it harder to apply either just a start time, or just a stop time, which I think could potentially be useful? 

What do we think about this possible API:
`def limit_df(df, fs, start, stop):`

If so, and whatever we choose, I think it should be matched in `limit_sig_times`, and if it keeps the 2 item list, maybe update the name. Maybe `limits`?

bycycle/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    return df
+
+
+def limit_sig_times(sig, times, xlim):
+    """Restrict signal and times to be within time limits.
+
+    Parameters
+    ----------
+    sig : 1d array
+        Time series to plot.
```
TomDonoghue, line 49: Just "Time series." Doesn't have to be to plot.

bycycle/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    df['sample_last_' + side_e] = df['sample_last_' + side_e] - int(fs * xlim[0])
+    df['sample_next_' + side_e] = df['sample_next_' + side_e] - int(fs * xlim[0])
+    df['sample_' + center_e] = df['sample_' + center_e] - int(fs * xlim[0])
+    df['sample_zerox_rise'] = df['sample_zerox_rise'] - int(fs * xlim[0])
+    df['sample_zerox_decay'] = df['sample_zerox_decay'] - int(fs * xlim[0])
+
+    return df
+
+
+def limit_sig_times(sig, times, xlim):
```
TomDonoghue, line 43: A bit pedantic, but this doesn't just limit the _times_, but rather the signal too. 
Maybe call is `limit_signal`?

bycycle/plts/features.py
```diff
@@ -0,0 +1,129 @@
...
+
+@savefig
+def plot_feature_catplot(df, burst_param, group_by=None, ax=None, **kwargs):
+    """ Plot a catplot for a cycle feature.
+
+    Parameters
+    ----------
+    df : pandas.DataFrame.
+        Dataframe output from :func:`~.compute_features`.
+    burst_param : str
```
TomDonoghue, line 75: the DF params aren't necessarily burst params, right? Should this just be called `param`, or maybe `label` or `param_label`?

bycycle/plts/features.py
```diff
@@ -0,0 +1,129 @@
...
+    ax.set_ylabel('# cycles', size=fontsize)
+
+    if xlim:
+        ax.set_xlim(xlim)
+
+    ax.hist(feature, bins=bins, color=color, alpha=alpha, **kwargs)
+
+
+@savefig
+def plot_feature_catplot(df, burst_param, group_by=None, ax=None, **kwargs):
```
TomDonoghue, line 68: I have a pedantic nitpick with the name `catplot`. It comes from seaborn, AFAIK, but I'm not sure it's such a general name. Also, since it means 'categorical plot', this name reads as 'plot feature categorical plot', which annoys me. 

I'm not so sure if there is a better name - but maybe something more general / agnostic, like `plot_feature_comparison`, or (since I think one could plot this without multiple categories) `plot_feature_data` or similar.

bycycle/plts/cyclepoints.py
```diff
@@ -0,0 +1,187 @@
...
+        plot_time_series(times, sig, colors=colors[0], ax=ax)
+
+    colors = colors[1:] if plot_sig is True else colors
+
+    plot_time_series(x_values, y_values, ax=ax, colors=colors, xlabel=xlabel,
+                     ylabel=ylabel, marker='o', ls='', **kwargs)
+
+
+@savefig
+def plot_cyclepoints_array(sig, fs, xlim=None, ps=None, ts=None, zerox_rise=None,
```
TomDonoghue, line 113: The API labels of `ps`, `ts`, `zerox_rise`, `zerox_decay` feels a little inconsistent to me. In other functions we've updated from `p` or `ps` to more explicitly say `peaks`, and since we don't say `extrema_peaks`, I'm not sure we need to note `zerox`. 

What about just calling them `peaks`, `troughs`, `rises`, `decays`, or something like that? To me, that feels the most consistent, and straight-forward. 

Also, I think I would put the `xlim` arg after all the cyclepoint ones.

bycycle/plts/cyclepoints.py
```diff
@@ -0,0 +1,187 @@
...
+                           zerox_decay=None, ax=None, **kwargs):
+    """Plot extrema and/or zero-crossings using arrays to define points.
+
+    Parameters
+    ----------
+    sig : 1d array
+        Time series to plot.
+    fs : float
+        Sampling rate, in Hz.fs
```
TomDonoghue, line 122: typo '.fs'

bycycle/plts/cyclepoints.py
```diff
@@ -0,0 +1,187 @@
...
+
+    df = limit_df(df, fs, xlim)
+    sig, times = limit_sig_times(sig, times, xlim)
+
+    # Extend plotting based on given arguments
+    x_values = []
+    y_values = []
+    colors = ['k']
+
+    if plot_extrema:
```
TomDonoghue, line 71: this function feels like it could do less work, given `plot_cyclepoints_array` exists. 

Why not:
```
peaks, troughs, rises, falls = None, None, None, None
if plot_extrema:
    # extract peaks & troughs from df and assign to variables
if plot_zerox:
    # extract rises & decays from df and assign to variables

plot_cyclepoints_array(sig, fs, peaks, troughs, rises, decays, **kwargs)
```

The only thing that might need to be updated / coordinated is the make consistent if & how `plot_cyclepoints_array` handles not plotting a signal (which is an option from the df func), but I definitely think we want to reduce what is otherwise copied code here.

TomDonoghue:
Okay @ryanhammonds - thanks for your work again updating this PR!

This PR is getting a little unwieldy, so I re-reviewed and tried to focus in on any potentially more important decisions / updates for defining the plotting API. Otherwise, things look good, and I mostly ignored any small nitpicks or inconsequential things. 

There are a small set of new comments, focused around a couple ideas:
- naming the `catplot` plot
- code duplication & API in in the cyclepoints plots
- code API in the new utils functions

I've tried to make suggestions on those elements, that we can also chat on. These updates potentially change the plotting API. Once they are addressed, I think we should call this PR done (finally!).

bycycle/utils.py
ryanhammonds: We can call rename it to `limit_signal`. I went with the `sig_times` because it limits and returns both.

bycycle/utils.py
ryanhammonds: Whoops, yeah, that is the order in neurodsp. I'll change this.

bycycle/utils.py
ryanhammonds: That was carried over over from plts.utils by accident. I'll change this.

bycycle/utils.py
TomDonoghue: Ohhh, ooops. 

I sorta mis-read it as "limit the times vector of the signal" rather than "limit sig & times". I see what the original name was now. 

I think I still vote for `limit_signal`, though, as that seems to me to be the goal / purpose (`times` is sorta just meta-data).

ryanhammonds: I'll change it to `limit_signal`. I think the original name could easily be misinterpreted.

bycycle/utils.py
ryanhammonds: I can break xlim into start/stop args. Do you think defaulting start to 0 and stop to -1 makes sense to allow only one variable to be defined? I'll need to change the code to work with -1.

TomDonoghue: My first thought was `(start=None, stop=None)`, and then they get applied individually, as like `if start: # limit start`, and `if stop: # limit stop`. I'm not sure start / stop as 0 / -1 would work very cleanly without more work, because the current values at time values, in seconds (right?), where a 0/-1 look like indices. Supporting both or defaulting to use indices seems like mixed convention / possibly confusing.

bycycle/plts/features.py
ryanhammonds: What about `plot_feature_categorical`, since it plots a feature given a categorical variable to group by? Otherwise, I like `plot_feature_comparison` over `plot_feature_data`.

bycycle/plts/cyclepoints.py
ryanhammonds: This function only uses the df from `compute_features` to determine cyclepoints. If someone first uses `compute_features`, cyclepoint arrays are not accessible. This is where I'd argue where a class structure may come in handy to carry around meta-data that may or may not be needed later. I've avoid this for now unless more cases like this come up.

To avoid recomputing the cyclepoints, the current implementation uses the df to determine cyclepoints instead. This is more efficient, computationally, than recomputing with `find_extrema` and `find_zerox`. If this function were to recompute these arrays, it could be dropped altogether, since the df will no longer be needed.

I should add the option for plotting the signal in `plot_cyclepoints_array`. The reason this is an option in the df functions, is for a case when `plot_bursts` is used first, and then cyclepoints points are overlayed. This was need to for the summary plotting function.

If copied code is a concern, I could move it to a private func that is used by both functions?

ryanhammonds: I replied but then I realized I misunderstood you point and deleted. This makes since to reduce code. :slightly_smiling_face:

bycycle/plts/features.py
TomDonoghue: I would vote for `plot_feature_comparison` over `plot_feature_categorical`. Although these are reasonably called 'categorical' plots, I think I'd say the 'category' is basically the feature, right? (as in, features are the 'categories' we have in our data). I'm happy with `plot_feature_comparison`

ryanhammonds: The feature is the continuous variable, not the categorical. The category is a group (patient vs control, etc) that a single feature is compared between. `plot_feature_comparison` kinda sounds like two features are going to be compared.

TomDonoghue: Ohhh right. Yeh, of course, you're totally right. 'Categorical' makes sense - but also this plot doesn't actually necessitate a category or comparison, right? As in `groupby` can be None, and it just plots on feature, no comparison across categories. Damnit, why is naming things so hard! I don't even know anymore, I'm going to cede this decision to you :) Call it whatever you think is best!

ryanhammonds:
Hi @TomDonoghue, I thought I responded after pushing updates related to your last review but it looks like I didn't. Whoops! This should be good to merge, if these changes look good to you.

TomDonoghue:
Awesome, I did another quick read / review of the code, and checked tests & docs locally!

Everything looks good to me! Thank you @ryanhammonds for a _ton_ of work on this one!

I'm merging in - ByCycle now officially has a plotting sub-module!

Issue #58, opened 2020-06-11
Bursts detection at the beginning of a signal


ryanhammonds:
Burst detection for the first 1-2 cycles can fail if too close to the beginning of a signal. This is because cycle points are sometimes missing and because amplitude/period consistency get np.nan values for the first cycle.  

Here is an example of the issue. The second signal is shifted 0.4 seconds and the first two cycles are missed.
![Screenshot_20200611_113437](https://user-images.githubusercontent.com/34786005/84426688-799f7380-abd8-11ea-83a9-c6d0baa2e43f.png)

ryanhammonds:
First/last cycle missing cyclepoints fixed in #59 and edges cycles for bursts throughout the signal fixed in #73.

Issue #47, opened 2020-04-08
Setting boundary values in find extrema function + inconsistent periods 


rahul-jayaram1:
Is there a way to manually set boundary values when running the find_extrema function? When I run it now, the function does not seem to calculate the extremas at the start and end of our signal. I ran the function's calculation code on its own using the input parameters as variables, and when I manually set the boundary value to 1, more of the extremas were detected than when I ran the function itself. I get an error when I try to assign the value of boundary as an input to the function. 

Another issue I have encountered is that when I obtain the period values from the returned dataframe of the compute_features function, these periods do not match with the differences in troughs of the plot of the signal with extremas highlighted. Also, sometimes the number of oscilations in the dataframe does not match the number of oscillations present on the plot of the signal with extremas highlighted.

nschawor:
Hi,
it would be easiest if you could provide a running code example (e.g. using synthetic data) with: the parameters you set, the output you get and the output you expect, then we can efficiently discuss it and see what the problem is here!

rahul-jayaram1:
Hi, 

I have set the following parameters. 
Fs = 600 
tlim = (-.1, .500) #time interval in seconds
tidx = np.logical_and(t>=tlim[0], t<tlim[1])
f_theta = (10,22)
f_lowpass = 100
N_seconds = 0.1
N_seconds_theta = 0.02

Our signal is a length 361 time series vector

here is the code I am running to get the output figure. 
signal_low= lowpass_filter(signal, Fs, f_lowpass, N_seconds=N_seconds, remove_edge_artifacts=False)
signal_narrow = bandpass_filter(signal, Fs, f_theta,
                                remove_edge_artifacts=False,
                                N_seconds=N_seconds_theta)
zeroriseN = _fzerorise(signal_narrow)
zerofallN = _fzerofall(signal_narrow)

Ps, Ts = find_extrema(signal_low, Fs, f_theta,
    filter_kwargs={'N_seconds':N_seconds_theta})
tidxPs = Ps[np.logical_and(Ps>tlim[0]*Fs, Ps<tlim[1]*Fs)]
tidxTs = Ts[np.logical_and(Ts>tlim[0]*Fs, Ts<tlim[1]*Fs)]

zeroxR, zeroxD = find_zerox(signal_low, Ps, Ts)

tidx = np.logical_and(t>=tlim[0], t<tlim[1])
[...]

PR #59, opened 2020-06-18
[BUG] Initial burst detection

ryanhammonds:
This is a fix for #58. It was also noted in #47. The first few cycles points were sometimes skipped if bursting occurred too close to the beginning of a signal. I made sure an example case worked, but more testing will need to be done to make sure there aren't other side effects.

@sydney-smith This branch should fix the issue you were having on the project you were working on.

This was the issue (bottom signal was sliced so the burst is closer to the beginning):

Screenshot_20200611_113437

And here are plots from this PR:

Screenshot_20200617_214307


ryanhammonds:
Just a note for the future: I need to make sure the changes here are stable for all start/end extrema combinations (i.e. starts and ends on peaks, starts and ends on troughs, starts on trough ends on peak, and starts on peak and ends on trough). This is because the enforcement of a peak start and trough end have been removed. A unit test ensuring this should be included before merging.

ryanhammonds:
I circled back around to this and found the issue to be caused by:
- Missing extrema points at the beginning and end of the signal.
- The default `boundary=None` (not 0, defaults to 1 cycle length of the lower f_range) in `find_extrema`.

Here is the example of the issue. These plots were made from master:
```
sig = np.load('tutorials/data/sim_bursting.npy')
fs = 1000 
f_alpha = (8, 12)
sig = filter_signal(sig, fs, 'lowpass', 30, n_seconds=.2, remove_edges=False)
sig_short = sig[int(.2*fs):]
```
full_orig
short_bound_orig
short_nobound_orig

The last two plots have the signal shortened, but contains all points where extrema should be placed. Default params (the second plot) cuts off the first two cycles. Setting `boundary=0` cuts off the first cycle.

Here is plot from the changes here:
short_nobound_new

The changes in the PR to produce this graph:
- Set the boundary to 0 in `compute_cyclepoints`. This could also be changed in `find_extrema`. 
- Pad the signal with zeros in `find_extrema` and then trim cyclepoints back to the original length of the signal.

The fix here is separate from marking the first and last cycle as non-burst in `detect_bursts_cycles`. That has not changed, but the issues above could results in the first 2-3 cycles missed as burst, rather than just the first.

TomDonoghue:
My initial thought: 
- right now, `find_extrema` defines default as `boundary=None` which (in code) then updates to 1 cycle. I don't think the preferred update is to change the value of `boundary` in `compute_cyclepoints` because that seems to me to define two different defaults. One way to think about it: this would mean that using `find_extrema` or `compute_cyclepoints` with default parameters could / would return different results, and I don't think that's expected behaviour
    - the question then: do we want the (consistent) default behaviour of boundary to be 0 points, or 1 cycle's worth of points. I'm not sure I have a great answer for this, not knowing how the original default was chosen. 

As of right now, I'm also not totally clear on how the padding reduces missed cyclepoints (over and above the changes the value of `boundary`). Basically, the slightly open decisions here are algorithmic (not strictly code). 

@ryanhammonds : can you perhaps add any info you know related to this, including:
- any thoughts or arguments you know of for / against any particular value for `boundary`
- add some notes on how the padding works to address the problem

Hopefully this info will help clarify some ideas of what to do. Relatedly, @sydney-smith - I think you mentioned this issue, so if you have any thoughts on your experience, and what you hope the algorithm _should_ do, then please let us know :)

TomDonoghue (commented):
Comments in main thread: I think there's a couple things to be sorted in terms of algorithm decisions, before reviewing the code implementation.

ryanhammonds:
@TomDonoghue 

> One way to think about it: this would mean that using find_extrema or compute_cyclepoints with default parameters could / would return different results, and I don't think that's expected behaviour 

I think the only results that would change is an added start/end row in the dataframe. It shouldn't affect cyclepoints in the middle of the signal.

>  - any thoughts or arguments you know of for / against any particular value for boundary

I don't see any issue with including cycles that are close to the beginning or end of a cycle and a boundary of zero won't affect later cycles either (at least from my quick observation from testing this; it may have other effects that I'm unaware of).

>  - add some notes on how the padding works to address the problem

The signal is filtered before determining cyclepoints, likely to prevent adding cyclepoints to very short rise/decays by smoothing the signal a bit. Padding with zero helps the start/end of the filtered signal reach sub/super zero values more easily, which is required for zerocrossing to be found. It's similar to natalie's comment in the meeting yesterday about reflecting the signal across zero. Reflecting is likely a more robust method than zero-padding, but it's may also be a little more computationally expensive. Both solutions feel like hacks, however.

TomDonoghue:
On find_extrema vs. compute_cyclepoints:
- yeh, not a big change. But I still don't see any reason to not have them default to the same (so, change the value in `find_extrema` rather than in `compute_cyclepoints`

On value of boundary:
- yeh, this makes sense to me. I actually guess the boundary may have been chosen to reflect filter edge artifacts, which means it relates to the next point. 

On filtering:
- Yeh, so I think the filtering is important to smooth things out a little. Zero-padding for filtering is fairly common (if a bit hacky). An example, is we actually do padding when we simulate filtered signals in NDSP, including calculating the precise length of data to pad to get the edge artifacts. See here:
https://github.com/neurodsp-tools/neurodsp/blob/master/neurodsp/sim/aperiodic.py#L217
This, I think, might well make sense to do here. I'm not sure if there would be any extra benefit to reflecting (which feels quite hacky). 
I don't think we want to skip filtering (change the algorithm) right now. That could be a future development idea, but let's aim for a solid 1.0 candidate, of the original algorithm(s) first. 
So I think, let's try adding padding, but make it optional (controlled by an argument). Maybe update to use the procedure from NDSP, which is a bit better for choosing the pad length. Then, let's test it out a bit, and if we're not sure, ask for some consults (or if anyone sees this and knows things - plz tell us the answer :))

Coda:
This actually relates back to comments from lab meeting. I think an idea we might want to develop in the documentation is that, when possible, it is better to run burst detection on continuous data, and then epoch. These edge things don't matter much in those cases, but they do become much more salient if it effects the beginning of every trial. Maybe we should up an example that demonstrates this approach, to also check if we want / need some extra utilities to support it.

sydney-smith:
I second that, @TomDonoghue. I think either the reflecting or zero-padding would work. Zero-padding for the equivalent of 2-3 cycles of the lower boundary of the burst range seems sufficient. Plus, in `find_extrema` the filter used to find zero-crossings is so narrow that most of the information in the signal is removed anyway so the tradeoff of computational expense for robustness doesn't really seem worth it.

The most important thing for my purposes is that the location of the first cycle is accurate, so I'm on board. 

One last concern (and this might be a totally different issue, possibly similar to #47) is that sometimes this "first cycle dropping" is happening even midway through the signal. I suspect this is because the `first_extrema` parameter in `find_extrema` is set automatically, causing the first half of a particularly sharp signal to be skipped. For example:

image

In this example, the burst is excluded because `first_extrema = 'peak'` by default. However, when I change the `center_extrema` parameter in `compute_features`, that first, very sharp cycle isn't excluded, even with the same burst_args dictionary.

image

I don't know how to approach this. How can the default first extrema selection be refined to optimally locate the first cycle?

ryanhammonds:
@TomDonoghue 
I'm on board with everything you've noted. I'll tag you back in after I've pushed updates.

@sydney-smith 
The `first_extrema` is default because when `center_extrema` is `trough`, the signal is inverted (so even though the `first_extrema` is forced to be `peak`, it's actually `troughs` relative to the un-inverted, original signal. I don't think changing this will help identify that first cycle as a burst.

Based on the plots, it looks like for the first cycle, `amplitude_consistency` is dropping to sub-threshold. This appears to be due to the increase in amplitude during that burst (i.e. it's less consistent as it goes from pre-burst to burst). That first cycle wouldn't be missed if the signal was reflected, but then you'd have an issue with the other side of the signal.

For consistency measures it makes sense to me to take a mean of running compute features in both directions (right-to-left and left-to-right). Right now, it's only right-to-left and I think that's the issue you've noted with that first cycle. But that would also require twice to computation. Maybe the optimal solution would be to have an argument that would allow lower consistency measure thresholds at the beginning of bursts?

ryanhammonds:
@TomDonoghue I pushed a commit based on your feedback:

- `boundary` set to zero from `find_extrema` rather than `compute_cyclepoints`
- optional `pad` argument (defaults to `True`) added to `find_extrema`
- the padding procedure updated to the same as edge-artifact prevention for filtered signals in ndsp

I also confirmed the plots I previously posted are the same (i.e. that zero-padding results in no missing cyclepoints).

bycycle/cyclepoints/extrema.py
```diff
@@ -51,13 +54,20 @@ def find_extrema(sig, fs, f_range, boundary=None, first_extrema='peak', filter_k
...
+    if pad:
+        filt_len = compute_filter_length(fs, 'bandpass', f_range[0], f_range[1],
+                                         n_seconds=filter_kwargs.get('n_seconds', None),
+                                         n_cycles=filter_kwargs.get('n_cycles', 3))
+
+    else:
+        filt_len = 0
+
+    sig_pad = np.zeros(len(sig) + filt_len)
+    sig_pad[int(filt_len):] = sig
```
TomDonoghue, line 67: So, I think right now, we only pad the beginning, to a full filter length?
Presumably we want to pad both ends of the signal?
Also, on one end, the padding only needs to be half a filter length. Once one gets half the filter length into the signal, the filter sees once full length data, and no longer has edge effects. 

So unless I'm missing something - the current length is good, but this should be updated to have the signal in the middle, with a half filter length buffer on either end.

bycycle/cyclepoints/extrema.py
```diff
@@ -51,13 +54,20 @@ def find_extrema(sig, fs, f_range, boundary=None, first_extrema='peak', filter_k
     if filter_kwargs is None:
         filter_kwargs = {}
 
-    # Default boundary value as 1 cycle length of low cutoff frequency
-    if boundary is None:
-        boundary = int(np.ceil(fs / float(f_range[0])))
+    # Pad beginning of signal with zeros to prevent missing cyclepoints
+    if pad:
```
TomDonoghue, line 58: This is minor / but I think this could be condensed a little, by doing the padding within this if. 
```
if pad:
    filt_len = compute_filter_length(...)
    sig = pass # line to pad the signal, without renaming it
```

One reason: we currently rename the variable to `sig_pad`, but it's actually not always padded. I think keeping it as `sig` is fine / makes more sense (we don't need a rename).

bycycle/cyclepoints/extrema.py
```diff
@@ -89,7 +99,11 @@ def find_extrema(sig, fs, f_range, boundary=None, first_extrema='peak', filter_k
         last_decay = decay_xs[t_idx]
         next_rise = rise_xs[rise_xs > last_decay][0]
         # Identify time of trough
-        troughs[t_idx] = np.argmin(sig[last_decay:next_rise]) + last_decay
+        troughs[t_idx] = np.argmin(sig_pad[last_decay:next_rise]) + last_decay
+
+    # Remove padding
```
TomDonoghue, line 104: I _think_ that instead of doing this down here, after the filter step, can't we just add:
```
if pad:
    sig, _ = remove_nans(sig)
```

If the padding is updated, as commented above, to pad both ends, the padding should line up exactly with the nan's added to the signal in `remove_filter_edges` (see: https://github.com/neurodsp-tools/neurodsp/blob/dec3b8167f7f0f402a8fc3bbfe14b87d94755d02/neurodsp/filt/utils.py#L211)

This way, the signal ends up padded, filtered, then cropped to expected around line 70, before it goes into the next procedures, and nothing needs to be fixed at the end.

TomDonoghue (changes requested):

TomDonoghue:
The planned updates all make sense to me! This is also something that I think we add, then get some user testing and exploring a little. 

In terms of the mechanics of the padding - I've left some comments. Unless I'm missing something to how it's working (very possible), I think there some updates we should do to how it's managed before merging.

bycycle/tests/cyclepoints/test_extrema.py
```diff
@@ -37,12 +37,12 @@ def test_find_extrema(first_extrema):
     minima = argrelextrema(sig, np.less)
 
     # Find peaks and troughs using bycycle and make sure match scipy
-    peaks, troughs = find_extrema(sig, fs, f_range, boundary=1, first_extrema=first_extrema)
+    peaks, troughs = find_extrema(sig, fs, f_range, first_extrema=first_extrema)
```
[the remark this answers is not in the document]
ryanhammonds: The simulated signal used in this test is a special case and this won't pass for other signals. `argrelextrema` + sig can find the ~~similar~~ exact same (`np.testing.assert_equal` will also pass) maximum/minima as `find_extrema` + sig_filt. The only difference is the scipy func returns two extra values, noted in your next comment (I'll need to look into this). `argrelextrema` will also not produce the same maxima/minima (as `find_extrema`) when the same filtered signal is used that is created within `find_extrema`,

For your second point, `find_extrema` will not work if only `sig_filt` is passed in. Rise/Decay points are found with sig_filt, but finding peaks and troughs requires the orginal signal:
```
sig_pad[int(filt_len):] = sig
# ...
peaks[p_idx] = np.argmax(sig_pad[last_rise:next_decay]) + last_rise
# ...
roughs[t_idx] = np.argmin(sig_pad[last_decay:next_rise]) + last_decay
```
I suppose `sig` and `sig_filt` could both be passed in, requiring `sig_filt` to be created prior to each `find_extrema` call. However, filtering can be turned off in the function (as is) using `filter_kwargs=None` and  `pad=False`, and full filtering flexibility/control could be achieved by adding `pass_type`, `remove_edges` to `find_extrema`.

bycycle/tests/cyclepoints/test_extrema.py
```diff
@@ -37,12 +37,12 @@ def test_find_extrema(first_extrema):
...
 
     if first_extrema == 'trough':
         assert len(peaks) == len(troughs)
         assert troughs[0] < peaks[0]
-        np.testing.assert_allclose(peaks, maxima[0])
+        np.testing.assert_allclose(peaks, maxima[0][:len(peaks)])
```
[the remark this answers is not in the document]
ryanhammonds: It's because when first_extrema is trough, the last extrema must be a peak and vice-versa. This will cause beginning/ending extrema to sometimes cut off peaks/troughs (i.e. first extrema is actually a peak, but `first_extrema` is set to `trough`, then the first actual peak (found with `argrelextrema`) will be missing, etc).

I've updated with comments explaining this.

bycycle/cyclepoints/extrema.py
ryanhammonds: I initially had it in my head the edge artifacts only occurred at the beginning of signals, but this isn't true. Thanks for the catch! Updated as you suggest (round `filt_len` up to the nearest even number if needed, and pad half that length on each side).

bycycle/cyclepoints/extrema.py
ryanhammonds: Agreed, condensed and the name is kept as `sig`. I also simplified using `sig = np.pad(sig, int(filt_len/2))`.

bycycle/cyclepoints/extrema.py
ryanhammonds: This ends up being a bit complicated. If the padding (two-sided) in `sig` and `sig_filt` are removed prior to the two `find_flank_zerox` calls, edge artifacts / missing extrema points will still occur. Here, the first and last peaks are missing when the padding/nans are removed prior the find_flank_zerox call. This is resolved by setting `remove_edges=False` when filtering, using the padded `sig` and `sig_filt` to compute peaks/troughs, and then removing/adjusting for padding on peaks/troughs.

Settings: `first_extrema=peak` and the signal starts with a peak and ends with a trough:

```
>           np.testing.assert_equal(peaks, maxima[0])
E           AssertionError: 
E           Arrays are not equal
E           
E           (shapes (98,), (100,) mismatch)
E            x: array([ 125,  225,  325,  425,  525,  625,  725,  825,  925, 1025, 1125,
E                  1225, 1325, 1425, 1525, 1625, 1725, 1825, 1925, 2025, 2125, 2225,
E                  2325, 2425, 2525, 2625, 2725, 2825, 2925, 3025, 3125, 3225, 3325,...
E            y: array([  25,  125,  225,  325,  425,  525,  625,  725,  825,  925, 1025,
E                  1125, 1225, 1325, 1425, 1525, 1625, 1725, 1825, 1925, 2025, 2125,
E                  2225, 2325, 2425, 2525, 2625, 2725, 2825, 2925, 3025, 3125, 3225,...
```

Here is the same signal / settings. The bottom plot shows the missing peaks when removing padding prior to `find_flank_zerox`.
Screenshot_20201015_124508

I'm not really sure why this is happening, but it seems better to leave as is.

bycycle/tests/cyclepoints/test_extrema.py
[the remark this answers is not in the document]
ryanhammonds: `pass_type` has been added as an optional kwarg. I didn't include `remove_edges` since this has to be done when padding anyways (it's actually done by adjusting `peaks` and `troughs` - the reason why is noted in a different comment.

ryanhammonds:
Hi @TomDonoghue, thanks for your helpful comments! I've addressed each one and pushed related updates. If everything looks good, this can be merged in and I can get a rc version released and figure out how to get the 1.0.0 docs to play nicely with the stable docs. Then it should be ready for user testing!

PR #60, opened 2020-06-23
mne interface example

alexrockhill:
It might be helpful to mne-python users to have an example of an interface with bycycle as in this example for a [pac toolbox](https://pactools.github.io/auto_examples/plot_mne_epoch_masking.html). I also added a bit out of curiosity using the pac simulation for both pac simulated as expected and "spurious" pac caused by spiking. The peak-trough asymmetry is lower in both the particular example of pac and spurious pac compared to a sine wave with pink noise while rise-decay asymmetry is closer to the same for all three but a bit skewed positive for spurious pac. I think this means it's working reasonably well? If the spurious pac spike amplitude gets increased, the same trend holds while it just looks like a spike train. I'm not entirely sure how to interpret this, maybe someone can comment.

A simple mne-python interface might be more useful in which case I can pare all this down significantly (or none at all in which case, totally fine to close without merging). I thought this might be useful and it's helped me understand the pac-asymmetry feature relationship a bit better. It would be great to know how to make it easier to understand/interpret though if anyone has any input.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,6 @@ notifications:
 # Install packages
 install:
   - pip install -r requirements.txt
-  - pip install -q mne
-  - pip install pactools
   - pip install pytest-cov
   - pip install codecov
   - pip install .
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ notifications:
 # Install packages
 install:
   - pip install -r requirements.txt
+  - pip install -q mne
+  - pip install pactools
   - pip install pytest-cov
   - pip install codecov
   - pip install .
--- a/examples/plot_mne_feature_distributions.py
+++ b/examples/plot_mne_feature_distributions.py
@@ -1,0 +1,224 @@
+"""
+MNE interface cycle feature distributions
+=============================================
+This example computes the distributions of cycle features using MNE objects
+"""
+
+####################################################################################################
+
+import mne
+import numpy as np
+import matplotlib.pyplot as plt
+import scipy.signal
+
+from pactools import simulate_pac, raw_to_mask, Comodulogram
+from pactools.utils.pink_noise import pink_noise
+from pactools.utils.validation import check_random_state
+
+from bycycle.features import compute_features
+
+####################################################################################################
+# Simulate and preprocess data
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+####################################################################################################
+
+
+def simulate_spurious_pac(n_points, fs, spike_amp=1.5, spike_fwhm=0.01,
+                          spike_fq=10., spike_interval_jitter=0.2,
+                          random_state=None):
+    """Simulate some spurious phase-amplitude coupling (PAC) with spikes
+
+    References
+    ----------
+    Gerber, E. M., Sadeh, B., Ward, A., Knight, R. T., & Deouell, L. Y. (2016).
+    Non-sinusoidal activity can produce cross-frequency coupling in cortical
+    signals in the absence of functional interaction between neural sources.
+    PloS one, 11(12), e0167351
+    """
+    n_points = int(n_points)
+    fs = float(fs)
+    rng = check_random_state(random_state)
+
+    # draw the position of the spikes
+    interval_min = 1. / float(spike_fq) * (1. - spike_interval_jitter)
+    interval_max = 1. / float(spike_fq) * (1. + spike_interval_jitter)
+    n_spikes_max = np.int(n_points / fs / interval_min)
+    spike_intervals = rng.uniform(low=interval_min, high=interval_max,
+                                  size=n_spikes_max)
+    spike_positions = np.cumsum(np.int_(spike_intervals * fs))
+    spike_positions = spike_positions[spike_positions < n_points]
+
+    # build the spike time series, using a convolution
+    spikes = np.zeros(n_points)
+    spikes[spike_positions] = spike_amp
+    # full width at half maximum to standard deviation convertion
+    spike_std = spike_fwhm / (2 * np.sqrt(2 * np.log(2)))
+    spike_shape = scipy.signal.gaussian(M=np.int(spike_std * fs * 10),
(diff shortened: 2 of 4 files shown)


alexrockhill:
peak-trough
rise-decay

The legend was only on the first one; blue is pac, orange is spurious pac and green is sine wave with pink noise.

TomDonoghue:
Hey @alexrockhill - thanks a lot for the PR! I think showing ByCycle with MNE objects, and also exploring PAC and using pactools would make for a really nice & useful example!

I was able to run your example, and have a quick look, at the output, and I do think the idea and general example would be useful to add. 

Some quick thoughts:
- At least on my current version of MNE, I had to update line 115, so I think this needs fixing:
    - from `raw = raw.filter(h_freq=30.)` to `raw = raw.filter(l_freq=None, h_freq=30.)`
- could you add a bit more narrative text to describe what is being done, and why? From my quick check, the logic makes sense, but I have a feeling a new user looking into our examples might be helped by a bit more guidance
    - relatedly, more annotations like adding titles to the comodulograms, and fixing them to the same colorbar scale might make it easier to follow
    - this could also include splitting up code sections a little more, and adding some more in line code comments
    - for a related example (since our currently ByCycle examples aren't super verbose either), if you want one, see this related example on the NeuroDSP package. 
        - https://neurodsp-tools.github.io/neurodsp/auto_examples/plot_mne_example.html
- It might be random generation, but I'm not sure the plots I'm getting are quite the expected output. For example, attached are what I generate as the signals and as the feature distributions. Perhaps most notably, I didn't see the difference in measured Peak-Trough symmetry. Do you know why? We can add a random-state setter to make the generation consistent, but also wanted to mention in case there is more to this variability. 

The time series I get:
Screen Shot 2020-06-24 at 2 20 15 PM
The feature distributions I get:
Screen Shot 2020-06-24 at 2 20 26 PM

^Are these plots as expected to you? I find the time series not super clear as to being able to see the differences (in the high frequencies), and the feature distributions aren't quite what I expected. 

I'm also tagging in @ryanhammonds, who is a maintainer of ByCycle. We are currently doing some updates to ByCycle, and Ryan can do a more detailed code check and might have some suggestions for the code, and/or for things like updating to use our new plot functions. 

@alexrockhill - for right now, can you respond to the points above? After that, Ryan can do a more specific code check, and either make any more suggestions that are needed, and/or (if you want) do direct edits to get it ready to merge in with the upcoming & updated version of ByCycle.

.travis.yml
```diff
@@ -12,6 +12,8 @@ notifications:
 # Install packages
 install:
   - pip install -r requirements.txt
+  - pip install -q mne
```
TomDonoghue, line 15: As a small point, these lines should be removed. Travis doesn't run the examples, and doesn't need these dependencies. If we add these examples, these requirements should be listed in `requirements-docs.txt`

alexrockhill:
@TomDonoghue thanks for the points, I was modeling it off of the other examples, but clearly should add more comments and will. I just pushed a version without the filter, I changed it right before pushing but should have tested it more because it throws things off. The example should work as expected now, my bad!

TomDonoghue:
Hey @alexrockhill - sorry for potential confusion, because I'm not sure if you noticed I updated my comment a bit after I first added it. 

What do you think about splitting out PAC & pactools from MNE usage? I don't want to get too far into any details before we settle on scoping & structure.

Otherwise, your edits & added comments so far look good - just a quick note to use left-hand side '#' for text sections, and not `""" """`, to render nice with sphinx (the tool that builds the docs). You can compare to other example or tutorial files.

alexrockhill:
@TomDonoghue, hey I saw the comment, I just hadn't gotten to it yet. Working on it some time this morning.

alexrockhill:
I checked all the channels not just the example channels I pushed and they are remarkable normal for asymmetries at the random ten second time window used; centered at 0.5 with a ~0.15 standard deviation. Not like the pac example.

Also any plans for adding Circle? That would make it easier to do the documentation formatting stuff.

ryanhammonds:
Hi @alexrockhill, thanks for the contribution! Circle as in circleci? We are currently using travis. It wouldn't hurt to add circleci, but I'm unaware of how it helps with documentation formatting.

alexrockhill:
I think as in [here](https://github.com/mne-tools/mne-python/pull/7863#issuecomment-639209352), you link the example plots to the comments.

ryanhammonds:
@alexrockhill I see, circleci artifacts looks helpful, but it may take a bit to get configured properly. In the meantime, you can build the site locally to checkout on formatting. Make sure to pull the most recent master branch, then:
```
pip install -e .
pip install -r requirements-docs.txt
cd doc
make clean
make html
```
This will create html files for the examples in  doc/_build/html/auto_examples.

alexrockhill:
> @alexrockhill I see, circleci artifacts looks helpful, but it may take a bit to get configured properly. In the meantime, you can build the site locally to checkout formatting. Make sure to pull the most recent master branch, then:
> 
> ```
> pip install -e .
> pip install -r requirements-docs.txt
> cd doc
> make clean
> make html
> ```
> 
> This will create html files for the examples in doc/_build/html/auto_examples.

Cool, just a suggestion.

ryanhammonds:
The last thing this PR needs is to update plotting to use the new `plts.features.plot_feature_hist` to decrease verbosity of mpl plotting calls. There may need to be a few other updates since we are in the process of a refactor and the api is slightly changing. I will directly push these updates sometime this week and then merge this PR. Thanks again for the contribution!

TomDonoghue:
Thanks a ton for the updates @alexrockhill! 

I'm going to have a more proper look through this later this week! As Ryan mentioned, there are a couple thing with code & plotting stuff, since we're currently updating the module, that Ryan can directly update - as long as that sounds good to you! We'll keep you in the loop and let you know if there is anything we need from you! 

Thanks a lot for the contributions, I'm happy to add coverage of more topics to the documentation!

alexrockhill:
Awesome :)

Issue #42, opened 2020-02-28
Warning while using compute_features


ShuangquanFeng:
c:\users\shuan\appdata\local\programs\python\python37-32\lib\site-packages\numpy\core\fromnumeric.py:2920: RuntimeWarning: Mean of empty slice.
  out=out, **kwargs)
c:\users\shuan\appdata\local\programs\python\python37-32\lib\site-packages\numpy\core\_methods.py:85: RuntimeWarning: invalid value encountered in double_scalars
  ret = ret.dtype.type(ret / rcount

I encountered this warning sometimes while using compute_features. Is that normal?

ryanhammonds:
Hi @ShuangquanFeng, I was able to reproduce this warning when running the unit test for `bycycle.burst.detect_bursts_cycles` and traced it back to the *monotonicity* calculation. The warning is caused by a short (length = 1) rise or decay period. In other words, when an adjacent peak/trough are separate by one sample.

The results is a nan value in the *monotonicity*  column of the dataframe output from `compute_features` and can safely be ignored.

ryanhammonds:
This should be resolved as #67 has been merged in. @ShuangquanFeng please reopen if this warning persists when using master, thanks!

PR #63, opened 2020-07-09
[ENH] Burst and feature refactor duplicate.

ryanhammonds:
I accidentally merge #61 into master, and it closed before review, without the possibility of reopening. I rolled the merged commits back so these commits could get a proper review before merging. Here is the original post:

This PR is related to #32 and #46. `bycycle.features` has been broken down into:

- `bycycle.features.shape`: computes cycle shape features (period, time_decay, volt_rise, etc)
- `bycycle.features.burst`: computes cycle burst features (monotonicity, amplitude_fraction, etc)
- `bycycle.features.features`: wraps these two functions to compute all features.

These functions may optionally return `df_samples`, which contains samples/signal indices for where cyclepoints occur. This should be the metadata that Erik's comment referred too.

Regarding #42, `bycycle.burst` now serves only to detect where a signal is oscillating, and to append the `is_burst` column to the feature dataframe. It doesn't require the signal as an arg (only needs the features dataframe) and won't require all features to be recomputed.

Here is an example, `compute_features` functions the same with exception of optionally returning a samples dataframe:
```
# Determine cycles that are bursting add an 'is_osc' column.
threshold_kwargs = {'amp_fraction_threshold': .5,
                    'amp_consistency_threshold': .5,
                    'period_consistency_threshold': .5,
                    'monotonicity_threshold': .5,
                    'n_cycles_min': 3}

# Compute both shape and burst features, and return cyclepoint sample indices in a separate df
df_features, df_samples = compute_features(sig, fs, f_range,  burst_method='cycles' \
            threshold_kwargs=threshold_kwargs, return_samples=True)
```

Tests have been updated. I'm holding off on updating the tutorials/examples until the code here has been finalized. I can add that to a separate PR to try to keep this PR smaller than the plotting refactor.

Also here is a list of variable names that have been renamed from 0.1.2:
```
# We decided on this since 'peaks' wasn't 'extrema_peaks'
zerox_rise -> rises
zerox_decay -> decays

# Consistency with neurodsp
x -> sig
Fs -> fs
N_cycles_min -> min_n_cycles

# Increased modularity
df -> df_samples, df_features

# Consistency between burst detection kwargs and dataframe column names
amplitude_fraction_threshold -> amp_fraction_threshold
amplitude_consistency_threshold -> amp_consistency_threshold

# Matching plotting kwargs with compute_features kwargs
osc_kwargs -> burst_kwargs

# pep8
Ps -> ps
Ts -> ts
zeroriseN -> rise_xs
zerofallN -> decay_xs
hilbert_increase_N -> deprecated
```


bycycle/cyclepoints.py
```diff
@@ -173,66 +140,64 @@ def find_zerox(sig, ps, ts):
...
+            flanks[idx] = extrema_start[idx] + int(np.median(find_flank_zero_xs(sig_temp, flank)))
 
-    return zerox_rise, zerox_decay
+    return flanks
 
 
-def extrema_interpolated_phase(sig, ps, ts, zerox_rise=None, zerox_decay=None):
-    """Use peaks (phase 0) and troughs (phase pi/-pi) to estimate instantaneous phase.
-    Also use rise and decay zero-crossings (phase -pi/2 and pi/2, respectively) if provided.
+def extrema_interpolated_phase(sig, peaks, troughs, rises=None, decays=None):
```
TomDonoghue, line 187: I stared at this function for a while, and I feel like there is a refactor available for it, I just didn't follow through on how. But I think it might be worth figuring out the logic of the doubled up code here, and at least document why there are two, and perhaps refactor.

bycycle/cyclepoints.py
```diff
@@ -257,26 +227,25 @@ def extrema_interpolated_phase(sig, ps, ts, zerox_rise=None, zerox_decay=None):
...
     pha_tpi = np.interp(times, times[~np.isnan(pha_tpi)], pha_tpi[~np.isnan(pha_tpi)])
     pha_tnpi = np.interp(times, times[~np.isnan(pha_tnpi)], pha_tnpi[~np.isnan(pha_tnpi)])
 
-    # For the phase time series in which the trough is negative pi:
-    # Replace the decaying periods with these periods in the phase time
-    # series in which the trough is pi
+    # For the phase time series in which the trough is negative pi, replace the decaying
+    #   periods with these periods in the phase time series in which the trough is pi
     diffs = np.diff(pha_tnpi)
     diffs = np.append(diffs, 99)
```
TomDonoghue, line 250: this `99` is a weird magic number for quite unclear reasons. I think we should figure out what this is / make a note.

bycycle/cyclepoints.py
```diff
@@ -292,3 +261,30 @@ def extrema_interpolated_phase(sig, ps, ts, zerox_rise=None, zerox_decay=None):
     pha_tnpi[-last_empirical_idx + 1:] = np.nan
 
     return pha_tnpi
+
+
+def find_flank_zero_xs(sig, flank):
```
TomDonoghue, line 266: So I refactored most of this file a bunch, and made sub-functions. But I didn't really solve the naming, as all the sub-functions ended up with sorta the same name (since they seem to have quite overlapping function). I think a ToDo is to logic through this file again, and check and probably update a naming scheme here.

bycycle/features/features.py
```diff
@@ -0,0 +1,146 @@
...
+import pandas as pd
+
+from bycycle.features.shape import compute_shape_features
+from bycycle.features.burst import compute_burst_features
+from bycycle.burst import detect_bursts_cycles, detect_bursts_df_amp
+
+###################################################################################################
+###################################################################################################
+
+def compute_features(sig, fs, f_range, center_extrema='peak', burst_detection_method='cycles',
```
TomDonoghue, line 12: for brevity, I think `burst_detection_method` could probably be `burst_method`. 
Likewise, and if so, then `burst_detection_kwargs` could probably be `burst_kwargs`.

bycycle/features/features.py
```diff
@@ -0,0 +1,146 @@
...
+        - ``time_rdsym`` : fraction of cycle in the rise period
+        - ``time_ptsym`` : fraction of cycle in the peak period
+        - ``band_amp`` : average analytic amplitude of the oscillation computed using narrowband
+          filtering and the Hilbert transform. Filter length is 3 cycles of the low cutoff
+          frequency. Average taken across all time points in the cycle.
+
+        When consistency burst detection is used (i.e. burst_detection_method == 'cycles'):
+
+        - ``amplitude_fraction`` : normalized amplitude
+        - ``amplitude_consistency`` : difference in the rise and decay voltage within a cycle
```
TomDonoghue, line 80: I think at some point `amp` got expanded in `amplitude` for feature names? I'm not sure we need this expansion - it makes those labels a fair bit longer, and for this case I think `amp` is a pretty clear short version.

bycycle/features/shape.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # Compute rise-decay symmetry features
+    shape_features['time_rdsym'] = shape_features['time_rise'] / shape_features['period']
+
+    # Compute peak-trough symmetry features
+    shape_features['time_ptsym'] = shape_features['time_peak'] / \
+            (shape_features['time_peak'] + shape_features['time_trough'])
+
+    # Compute average oscillatory amplitude estimate during cycle
+    amp = amp_by_time(sig, fs, f_range, hilbert_increase_n=hilbert_increase_n, n_cycles=3)
```
TomDonoghue, line 135: Here, n_cycles should use a controllable input, I think? Maybe from `find_extrema_kwargs`?

bycycle/features/shape.py
```diff
@@ -0,0 +1,213 @@
...
+        df_features['time_rdsym'] = 1 - df_features['time_rdsym']
+        df_features['time_ptsym'] = 1 - df_features['time_ptsym']
+
+    if return_samples:
+        return df_features, df_samples
+
+    return df_features
+
+
+def compute_samples(peaks, troughs, decays, rises):
```
TomDonoghue, line 175: This function sounds like it should compute all the cyclepoints - but it seems like it mostly just collects them. 

Any reason to not make this function do the computing as well? As it gets called from the above function, then the following lines could be adopted into this function:

```
# Find extrema and zero-crossings locations in the signal
peaks, troughs = find_extrema(sig, fs, f_range, **find_extrema_kwargs)
rises, decays = find_zerox(sig, peaks, troughs)
```

The API would be updated to it to work accordingly, to something like:
`def compute_samples(sig, fs, f_range, **extram_kwargs)`

bycycle/features/shape.py
```diff
@@ -0,0 +1,213 @@
...
+    # Compute average oscillatory amplitude estimate during cycle
+    amp = amp_by_time(sig, fs, f_range, hilbert_increase_n=hilbert_increase_n, n_cycles=3)
+    shape_features['band_amp'] = [np.mean(amp[troughs[sig_idx]:troughs[sig_idx + 1]]) for sig_idx in
+                                  range(len(df_samples['sample_peak']))]
+
+    # Convert feature dictionary into a DataFrame
+    df_features = pd.DataFrame.from_dict(shape_features)
+
+    # Rename columns if they are actually trough-centered
+    if center_extrema == 'trough':
```
TomDonoghue, line 143: This feels like it could be a helper / sub-function

bycycle/features/shape.py
```diff
@@ -0,0 +1,213 @@
...
+    peaks, troughs = find_extrema(sig, fs, f_range, **find_extrema_kwargs)
+    rises, decays = find_zerox(sig, peaks, troughs)
+
+    # For each cycle, identify the sample of each extrema and zero-crossing
+    df_samples = compute_samples(peaks, troughs, decays, rises)
+
+    # Initialize dictionary for shape features
+    shape_features = {}
+
+    # Compute durations of period, peaks, and troughs
```
TomDonoghue, line 110: These are all very small calculations, but it means this function is calculating a lot of different things overall, and there is no way to compute each measure independently. I think maybe some (all?) of these computation should be implemented as sub-functions, that operate on arrays, that get called from this function. 

These would be very simple array functions, doing quick calculations, but I think it would help with the modularity a bit. 
For example:
`def compute_distance(vals1, vals): return vals1 - vals2`
Could be used to compute durations, etc.

TomDonoghue:
Okay, this one has a slightly dis-jointed review again. 

I looked through mostly on a local copy of code, so a lot of small fixes and so on, I just edited as I went. You can check these in my 3rd commit. I dug in a bit more on the `cyclepoints` file, playing with some refactors. Please sanity check these in the 2nd commit. And then, overall, I collected some thoughts and suggestions and left individual review comments on leftover things. 

Overall, all looks really nice! At some point before full 1.0 release, I'll need to play with some data, and make sure the overall workflow and logic make sense - so far I've focused on what comes to mind as I checked file by file, etc.

bycycle/burst.py
```diff
@@ -59,90 +55,65 @@ def detect_bursts_cycles(df, sig, amp_fraction_threshold=0.,
...
-    df = _min_consecutive_cycles(df, n_cycles_min=n_cycles_min)
-    df['is_burst'] = df['is_burst'].astype(bool)
+    df_features = _min_consecutive_cycles(df_features, n_cycles_min=n_cycles_min)
+    df_features['is_burst'] = df_features['is_burst'].astype(bool)
 
-    return df
+    return df_features
 
 
 def _min_consecutive_cycles(df_shape, n_cycles_min=3):
```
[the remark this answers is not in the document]
ryanhammonds: Updated to return arrays.

bycycle/features/burst.py
```diff
@@ -0,0 +1,250 @@
...
+
+import numpy as np
+import pandas as pd
+
+from neurodsp.burst import detect_bursts_dual_threshold
+
+###################################################################################################
+###################################################################################################
+
+def compute_burst_features(df_shape_features, df_samples, sig, dual_threshold_kwargs=None):
```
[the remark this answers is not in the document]
ryanhammonds: I agree, this felt strange. Across all functions, the kwargs have been updated (24dd493):

- burst_method : either 'cycles' or 'amp'
- burst_kwargs : additional 'amp' args (not required for 'cycles')
- threshold_kwargs : feature thresholds

bycycle/features/burst.py
```diff
@@ -0,0 +1,250 @@
...
+        decay_mono = np.mean(np.diff(decay_period) < 0)
+        rise_mono = np.mean(np.diff(rise_period) > 0)
+        monotonicity[idx] = np.mean([decay_mono, rise_mono])
+
+    return monotonicity
+
+
+def compute_burst_fraction(df_samples, sig, fs, f_range, amp_threshes=(1, 2),
+                           n_cycles_min=3, filter_kwargs=None):
+    """Compute the proportion of a cycle that is bursting.
```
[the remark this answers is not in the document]
ryanhammonds: I think this is correct but I reworded to hopefully make it clearer (829b84d). The burst fraction is computed for each cycle:

```
fraction_bursting = np.mean(is_burst[int(row['sample_last_' + side_e]):
                                     int(row['sample_next_' + side_e] + 1)])
```

Where `is_burst` is a boolean version of the bool array returned from neurodsp's `detect_bursts_dual_threshold`.
The line above takes the mean of each cycle's portion of this binary array. For example, if a burst has three samples,
and the ndsp func gives np.array([True, True, False]) for that section of the signal, the burst_fraction is 0.66 for that cycle.

Older versions of bycycle required the burst_fraction to be 1, but I've allowed this to be overwritten
by passing a float into the `burst_fraction_threshold` kwarg of the `detect_bursts_amp` function.

bycycle/cyclepoints.py
ryanhammonds: I refactor this (3c42063). I don't think there is any duplicate code here. pha_tpi and pha_npi have inverted signs and are merged together. This is so the returned phase is always rising. The function I added does the merging.

bycycle/cyclepoints.py
ryanhammonds: 99 is added to keep the length of the interpolated phase array == length of the signal. The negative pi phase array is always used for the last sample. I replaced 99 with np.nan and left a comment (3c42063).

bycycle/cyclepoints.py
ryanhammonds: It was difficult to find names to differenitate them. Updated in 75d9945. These aren't great still, however.

`find_flank_zero_xs -> find_flank_zerox`
I renamed this to keep it consistent with the other zerox funcs.

`_find_zerox_flanks -> _find_flank_midpoints`
I renamed this to differentiate it from other zerox funcs. This function finds location of the median
voltage between two extrema, rather than where the voltage == 0 as in find_flank_zerox.

After renaming this, I guess `find_zerox` is kind of a misnomer since it only returns zero-crossings
for cycles that are perfectly symetrical, which is unlikely in real data.

bycycle/features/features.py
ryanhammonds: Renamed as you suggested (a5f4ed8).

bycycle/features/features.py
ryanhammonds: Renamed as you suggested (a5f4ed8).

bycycle/features/shape.py
ryanhammonds: Fixed this (a953843).

bycycle/features/shape.py
ryanhammonds: This is a good idea, updated (da3fd5c).

bycycle/features/shape.py
ryanhammonds: Updated (984f578).

bycycle/features/shape.py
ryanhammonds: I broke these down in to sub functions (f31f24c).

ryanhammonds: I broke these down in to sub functions (f31f24c).

TomDonoghue:
I've had another check through the code. All the updates look great!

I did another sweep through the code, and made some small lint related updates and doc fixes, etc. And then I was playing with organizing 

Things I updated:
- consolidated on the parameter name `min_n_cycles` (it varied a little in different places)
- drop access to `hilbert_increase_n` (see: https://github.com/neurodsp-tools/neurodsp/issues/214)
- re-org individual files into sub-folders, and related re-orgs for this. 
    - This was done partly for a hopefully clear conceptual layout, and to be forward looking (not everything needed spitting out _now_, but hopefully this is a good layout to develop on). 
    - Note that generally access to functions (how they get imported) didn't change much, because most things are accessible through `__init__`. The organization is more internal than for the user. 

So, @ryanhammonds I'm going to throw it back to you to review the updates. Some tasks:
- Review the updates. Do my tweaks make sense, and what do you think of the updated organizations?
- I didn't properly update the tests. Can you update the test organization to match the folder structure and file layout, etc.
- Some values, such as in `detect_bursts_cycles` have ranges that the values should take, but we never check them. Can you add a checker for this? It could be modelled on what we added to NeuroDSP in here (https://github.com/neurodsp-tools/neurodsp/pull/210)
- Can you update your original comment in this thread, to be up to date with the new names & API, etc? This will help for things like the Changelog later.

After these updates, I think we can do one final check here, and then hopefully we are done on the main refactor here!

ryanhammonds:
Hi @TomDonoghue, your updates and restructure make sense to me. The one refactor that maybe could be renamed is `bycycle.cyclepoints.extrema` to `bycycle.cyclepoints.cyclepoints`, since this sub-module includes zero-crossings in addition to extrema and a similar naming was kept in `bycycle.features.features`. I'm not sure if naming a folder and submodule the same is normal. If not, maybe both could be renamed. Other than that, everything looks great!

I push commits to restructure tests and to add argument checking anywhere where an expected range is known. I think this PR is ready to merge.

TomDonoghue:
Okay @ryanhammonds - I think we are there!

I made some mild updates:
- checking & organizing tests a bit. for folders of tests, adding `__init__` avoids issues of having test files with the same names. 
- updating naming within `cyclepoints`. since there are two different kinds of tasks, i split up into extrema & zerox. 
- update a little checking. Basically, I recommend we check a value in the function where it's defined / used - otherwise we can end up checking the same value multiple times, which is also harder to maintain. Related to this, I removed some checks, like for `f_range`, which gets checked when it's passed through to the filter function
- general lints & words stuff.

TomDonoghue:
For the sake of moving on, and our sanity with this monstrous PR - I'm going to merge this in now. 

Any leftover nitpicks with this refactor can be addressed in subsequent updates, as we finish 1.0. 

Thanks for a ton of work on this one @ryanhammonds!

PR #65, opened 2020-07-27
Add requirements.txt to MANIFEST.in

synapticarbors:
It looks you can't build this package from source since the `setup.py` fails since the `requirements.txt` isn't included in the source distribution. This PR adds the file to the `MANIFEST.in`

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include LICENSE
+include requirements.txt


synapticarbors:
Also just to be clear:

```
$ pip install --no-binary=:all: bycycle

ERROR: Command errored out with exit status 1:
...
Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/ld/gx_9s3_n55g1dkhj59zjlxww0000gn/T/pip-install-nbu1vzpc/bycycle/setup.py", line 15, in <module>
        with open("requirements.txt") as requirements_file:
    FileNotFoundError: [Errno 2] No such file or directory: 'requirements.txt'
```

TomDonoghue:
Thanks @synapticarbors for catching this issue and the help / comments in the linked conda-forge PR for explaining this issue. This was our bad on not packing the `requirements.txt` file when we started using it from `setup.py`. 

Fix looks good - merging in!

PR #67, opened 2020-08-12
[BUG] Monotonicity Warning

ryanhammonds:
This is related to #42. The issue involved warnings when computing monotonicity. This occurred when `sample_last_trough` and `sample_peak` (or `sample_peak` and `sample_next_trough` for decay periods) are adjacent. This should be fixed be adding one to the end slice index. The `rise_period` should include the start/end point, otherwise the segments directly to the left/right of the peak are left out when taking the np.diff of the rise and decay periods. Here is sample code to reproduce the warning:

```
import numpy as np
from neurodsp.filt import filter_signal
from bycycle.plts import plot_burst_detect_summary
from bycycle.features import compute_features

sigs = np.load('tutorials/data/sim_experiment.npy')
sig = sigs[0]
fs = 1000
sig = filter_signal(sig, fs, 'lowpass', 30, n_seconds=.2, remove_edges=False)

threshold_kwargs = {'amp_fraction_threshold': .2,
                    'amp_consistency_threshold': .5,
                    'period_consistency_threshold': .5,
                    'monotonicity_threshold': .8,
                    'min_n_cycles': 3}

df_features, df_samples = compute_features(sig, fs, (7, 13), threshold_kwargs=threshold_kwargs)

# Plot the cycle that raises the warning
print(f'Samples to seconds: \n{df_samples.iloc[25] / fs}\n')

plot_burst_detect_summary(df_features, df_samples, sig, fs, threshold_kwargs,
                          plot_only_result=True, xlim=(2.768, 2.8))


print(f'{df_samples.iloc[25]}\n')
rise_period = sig[2769:2770]
print(rise_period) # diff of a single value is an empty array

rise_mono = np.mean(np.diff(rise_period) > 0) # mean of an empty array raises the warning
```

### DIFF
--- a/bycycle/features/burst.py
+++ b/bycycle/features/burst.py
@@ -200,12 +200,12 @@ def compute_monotonicity(df_samples, sig):
     for idx, row in df_samples.iterrows():
 
         if 'sample_peak' in df_samples.columns:
-            rise_period = sig[int(row['sample_last_trough']):int(row['sample_peak'])]
-            decay_period = sig[int(row['sample_peak']):int(row['sample_next_trough'])]
+            rise_period = sig[int(row['sample_last_trough']):int(row['sample_peak'])+1]
+            decay_period = sig[int(row['sample_peak']):int(row['sample_next_trough'])+1]
 
         else:
-            decay_period = sig[int(row['sample_last_peak']):int(row['sample_trough'])]
-            rise_period = sig[int(row['sample_trough']):int(row['sample_next_peak'])]
+            decay_period = sig[int(row['sample_last_peak']):int(row['sample_trough'])+1]
+            rise_period = sig[int(row['sample_trough']):int(row['sample_next_peak'])+1]
 
         decay_mono = np.mean(np.diff(decay_period) < 0)
         rise_mono = np.mean(np.diff(rise_period) > 0)
--- a/bycycle/features/burst.py
+++ b/bycycle/features/burst.py
@@ -131,17 +131,31 @@ def compute_amp_consistency(df_shape_features, df_samples):
 
     for cyc in range(1, cycles-1):
 
-        consist_current = np.min([rises[cyc], decays[cyc]]) / np.max([rises[cyc], decays[cyc]])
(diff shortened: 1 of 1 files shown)


ryanhammonds:
I also added a warning suppression for a division by zero that can occur in `compute_amp_consistency` (`bycycle.features.burst`).

TomDonoghue:
Okay, so I see this as two slightly different things:
- 1) computing monotonicity should include the actual peak / trough values
- 2) in the edge case of adjacent peak / troughs, monotonicity is undefined (nan)

On 1 - I agree, in theory. In practice, the peak / trough values are chosen as argmax / argmin, right? So there will be, by definition, the expected increase / decrease between the extrema and adjacent points, I think. So measure the monotonicity of these points is not actually informative. 
That said, it still might still be practically useful & more consistent to compute including peaks & troughs, and I don't think I see a reason why not. So... yeh, let's add that? The only note is that this may slightly change the results for some things, as having an extra monotonicity value that is always monotonic will globally increase all these measures. I don't _think_ that's a problem per se - it seems consistent that the last point in a cycle would be included, even if it's technically uninformative. 

On 2 - well, if peak & trough are next to each other, the cycle is nonsense really. So I'm not so worried about the nans (they are, perhaps, the most appropriate option). If the occurrence of undefined measures is reduced by changes in 1, that seems fine. 

So, based on my understandings on all this, I think, in too many words, this all seems good? In terms of the actual code to do the changes, as long as we are all still happy with the proposed changes, the code to do so looks good to me.

TomDonoghue (approved):

ryanhammonds:
For 1: I agree with everything you noted. I don't think it's totally uninformative though as it won't have the same effect on all rise/decay periods, dependent on how long the rise/decay periods are, since the mean is taken. Anyways, we are on the same page!

For 2: Agreed.

I'll merge this in soon!

TomDonoghue (commented):
I think something has gone weird with notifications, and it's still listing this PR in my list of 'review requests', so I'm "reviewing" it....

PR #69, opened 2020-08-15
[ENH] Docstring examples

ryanhammonds:
The first commit is the open PRs squash into one, which can be dropped as the open PRs are merged in. The second commit has the added examples.

### DIFF
--- a/bycycle/burst/amp.py
+++ b/bycycle/burst/amp.py
@@ -25,15 +25,15 @@ def detect_bursts_amp(df_features, burst_fraction_threshold=1, min_n_cycles=3):
 
     Examples
     --------
-    Apply thresholding for dual amplitude burst detection.
+    Apply thresholding for dual amplitude burst detection:
 
     >>> from bycycle.features import compute_burst_features, compute_shape_features
     >>> from neurodsp.sim import sim_bursty_oscillation
-    >>> fs, f_range = 500, (8, 12)
-    >>> sig = sim_bursty_oscillation(10, fs, 10)
-    >>> df_shapes, df_samples = compute_shape_features(sig, fs, f_range)
+    >>> fs = 500
+    >>> sig = sim_bursty_oscillation(10, fs, freq=10)
+    >>> df_shapes, df_samples = compute_shape_features(sig, fs, f_range=(8, 12))
     >>> df_burst = compute_burst_features(df_shapes, df_samples, sig, burst_method='amp',
-    ...                                   burst_kwargs={'fs': fs, 'f_range': f_range})
+    ...                                   burst_kwargs={'fs': fs, 'f_range': (8, 12)})
     >>> df_burst = detect_bursts_amp(df_burst)
     """
 
--- a/bycycle/burst/amp.py
+++ b/bycycle/burst/amp.py
@@ -22,6 +22,19 @@ def detect_bursts_amp(df_features, burst_fraction_threshold=1, min_n_cycles=3):
     -------
     df_features : pandas.DataFrame
         Dataframe updated, with a additional column to indicate if the cycle is part of a burst.
+
+    Examples
+    --------
+    Apply thresholding for dual amplitude burst detection.
+
+    >>> from bycycle.features import compute_burst_features, compute_shape_features
+    >>> from neurodsp.sim import sim_bursty_oscillation
+    >>> fs, f_range = 500, (8, 12)
+    >>> sig = sim_bursty_oscillation(10, fs, 10)
+    >>> df_shapes, df_samples = compute_shape_features(sig, fs, f_range)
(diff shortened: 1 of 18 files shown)


bycycle/burst/utils.py
```diff
@@ -20,6 +20,13 @@ def check_min_burst_cycles(is_burst, min_n_cycles=3):
     -------
     is_burst : 1d array
         Updated burst array.
+
+    Examples
+    --------
+    Removed bursts with less than 3 consectutive cycles.
+
+    >>> is_burst = np.random.choice([True, False], size=10)
```
TomDonoghue, line 28: I think the random array is potentially a little confusing here. 
We should be able to use an exact array, and then show the answer, right?

Like:

```
>>> is_burst = np.array([False, True, True, False, True, True, True, True, False, False ])
>>> is_burst = check_min_burst_cycles(is_burst)
np.array([False, False False, False, True, True, True, True, False, False ])

File: bycycle/burst/cycle.py [hunk 2]
```diff
@@ -62,6 +62,18 @@ def detect_bursts_cycles(df_features, amp_fraction_threshold=0., amp_consistency
...
       used.
+
+    Examples
+    --------
+    Apply thresholding for consistency burst detection.
+
+    >>> from bycycle.features import compute_burst_features, compute_shape_features
+    >>> from neurodsp.sim import sim_bursty_oscillation
+    >>> fs, f_range = 500, (8, 12)
+    >>> sig = sim_bursty_oscillation(10, fs, 10)
```
TomDonoghue, line 73: for clarity on what the inputs are, I think I would write out `freq=10` in the inputs to sim_bursty_oscillation. 
also, since `f_range` is only used once, it could be written in the inputs to the func below (as `f_range=(8, 12)`), then we don't need double assignment on a line.

bycycle/burst/amp.py
```diff
@@ -22,6 +22,19 @@ def detect_bursts_amp(df_features, burst_fraction_threshold=1, min_n_cycles=3):
     -------
     df_features : pandas.DataFrame
         Dataframe updated, with a additional column to indicate if the cycle is part of a burst.
+
+    Examples
+    --------
+    Apply thresholding for dual amplitude burst detection.
+
+    >>> from bycycle.features import compute_burst_features, compute_shape_features
```
TomDonoghue, line 30: I made comments on organizing args in `burst/cycle` below. Not too important how you choose to organize them, but if you do update, please check through and make consistent in all examples that have the same layout (importing and using sim_bursty_oscillation, etc).

bycycle/burst/utils.py
```diff
@@ -20,6 +20,13 @@ def check_min_burst_cycles(is_burst, min_n_cycles=3):
     -------
     is_burst : 1d array
         Updated burst array.
+
+    Examples
+    --------
+    Removed bursts with less than 3 consectutive cycles.
```
TomDonoghue, line 26: tense nitpick, but I think it should be:
`Remove bursts with less than 3 consectutive cycles:`

Present tense, and with trailing colon, like other docstrings.

bycycle/features/burst.py
```diff
@@ -54,6 +54,18 @@ def compute_burst_features(df_shape_features, df_samples, sig,
         When dual threshold burst detection is used (i.e. burst_method == 'amp'):
 
         - ``burst_fraction`` : fraction of a cycle that is bursting
+
+    Examples
+    --------
+    Determine burst features.
```
TomDonoghue, line 60: This might just be me, but 'determine' feels like a slightly odd word here. 
Maybe 'Compute'?

bycycle/group/features.py
```diff
@@ -0,0 +1,276 @@
...
+
+    >>> import numpy as np
+    >>> from neurodsp.sim import sim_bursty_oscillation
+    >>> fs, f_range = 500, (8, 12)
+    >>> sigs = np.array([sim_bursty_oscillation(10, fs, 10) for i in range(10)])
+    >>> compute_kwargs = {'burst_method': 'amp', 'threshold_kwargs':{'burst_fraction_threshold': 1}}
+    >>> df_features = compute_features_2d(sigs, fs, f_range, return_samples=False, n_jobs=2,
+    ...                                   compute_features_kwargs=compute_kwargs)
+
+    Compute the features of a 2d array in parallel using unique compute_features kwargs to examine
```
TomDonoghue, line 67: I'm not sure `unique compute_features kwargs` is particularly obvious, so maybe "using individualized settings per signal" or something similar.

TomDonoghue:
Some as the doc PR - I had a quick check, and made some small comments. Once everything is merged in, this looks good to me to merge in, to up the doc-example coverage.

ryanhammonds:
I addressed all of your comments in the latest commit, added a few missing docstring examples, and updated to ensure `make doctests` all passed.

Issue #62, opened 2020-07-06
DOC: Add a glossary


TomDonoghue:
In [FOOOF](https://fooof-tools.github.io/fooof/glossary.html) and [NDSP](https://neurodsp-tools.github.io/neurodsp/glossary.html) we have glossary pages on the documentation sites that explain the key terminology and words we use. 

ByCycle doesn't yet have one - but I think we should add one. Once we settle on the new 1.0 API, let's revisit and collect / define the key term we need and use across the module. We can use this issue to collect key terms and definitions, etc.

TomDonoghue:
At some point in the past when I did some updates I started collected some terms for this, so dropping in here. 

This list is incomplete, and any definition are copied randomly from places / may not be good.

```
sinusoidality
    xx
period
    A single cycle of a rhythm, defined as the time between two consecutive troughs (or peaks).
rise-decay symmetry
    The fraction of the period in the rise phase.
peak-trough symmetry
    The fraction of the period in the peak phase.
period
    The time between consecutive troughs (or peaks, if default is changed).
amplitude
    Average voltage changes of the rise and decay
amplitude consistency
    The amplitude consistency of a cycle is equal to the maximum relative difference between rises and 
    decay amplitudes across all pairs of adjacent rises and decays that include one of the flanks in the 
    cycle (3 pairs) (e.g. if a rise is 10mV and a decay is 7mV, then its amplitude consistency is 0.7)
period consistency
    Period consistency is equal to the maximum relative difference between all pairs of adjacent periods
    that include the cycle of interest (2 pairs: current + previous cycles and current + next cycles) (e.g. if the
    previous, current, and next cycles have periods 60ms, 100ms, and 120ms, respectively, then the period 
    consistency is min (60/100, 100/120) = 0.6.))
monotonicity
    The monotonicity is the fraction of samples that the instantaneous derivative (numpy.diff) is consistent with 
    the direction of the flank. (e.g. if in the rise, the instantaneous derivative is 90% positive, and in the decay, 
    the instantaneous derivative is 80% negative, then the monotonicity of the cycle would be 0.85 ((0.9+0.8)/2). 
    The rise and decay flanks of the cycle should be mostly monotonic. 
```

ryanhammonds:
Completed in #70

PR #70, opened 2020-08-17
[DOC] Site updates

ryanhammonds:
This PR depends on the other open PRs. Once they are merge, the first commit may be dropped. The second and third commits are fixes for issues I noticed when going through the site. 

The tutorials commit  (08ac29e) updates pre-existing tutorials to use updated funcs from the refactor. Once the other PRs are merged in, we can discuss new tutorials/examples that would be helpful to add. I think a short tutorial on processing 2d/3d arrays would be useful. It could expand on the related docstring examples in #69. 

A glossary was also added to address #62 in f3e6e39.

Lastly, I fixed intersphinx linking to neurodsp, which should close out voytekresearch/VoytekLabPrivate#7.


doc/glossary.rst
```diff
@@ -0,0 +1,85 @@
...
+    Peak-trough symmetry (``time_ptsym``)
+        The fraction of the period in the peak phase.
+
+    Sinusoidality
+        A symmetrical wave with 0.5 rise-decay and peak-trough symmetry.
+
+    Voltage Peak (``volt_peak``)
+        The voltage at the current peak.
+
+    Voltage Trough (``volt_trough``)
```
TomDonoghue, line 40: typo on pervious - but also I think the word should be 'current'?

doc/glossary.rst
```diff
@@ -0,0 +1,85 @@
...
+Shape
+-----
+Features that describe the shape of a cycle.
+
+.. glossary::
+
+    Period (``period``)
+        A single cycle of a rhythm, defined as the time between two consecutive troughs (or peaks).
+
+    Peak (``time_peak``)
```
TomDonoghue, line 16: this feels a slightly odd definition for peak / trough, and I think _not_ how they are computed. 
The peak / trough values are computed as the max / min values of detected cycles (where cycles are operationally defined as 'time between zero-crossings'. 

I think 'time between' doesn't quite capture this, and is slightly different from the intuitive idea of 'max' / 'min'.

So, maybe something like:
```
Peak
    The point of maximum height of a cycle. 
```

Does this track?

doc/glossary.rst
```diff
@@ -0,0 +1,85 @@
...
+
+Burst
+-----
+Additional shape features that aid in determining where a signal may be bursting.
+
+.. glossary::
+
+    Amplitude Fraction (``amp_frac``)
+        The average amplitude, relative to all other cycles. A value of 1.0 represents the cycle with
+        the maximum average amplitude. While values approaching 0.0 represents the minimum.
```
TomDonoghue, line 63: last sentence has phrasing issue - I think no 'while'.

doc/glossary.rst
```diff
@@ -0,0 +1,85 @@
...
+        consistency is min(60/100, 100/120) = 0.6)).
+
+    Monotonicity (``monotonicity``)
+        The monotonicity is the fraction of samples that the instantaneous derivative (numpy.diff) is consistent with
+        the direction of the flank. (e.g. if in the rise, the instantaneous derivative is 90% positive, and in the decay,
+        the instantaneous derivative is 80% negative, then the monotonicity of the cycle would be 0.85 ((0.9+0.8)/2)).
+        The rise and decay flanks of the cycle should be mostly monotonic.
+
+    Burst Fraction (``burst_fraction``)
+        The proportion of a cycle's samples that are bursting according to the dual amplitude threshold algorithm
```
TomDonoghue, line 83: is this always / only according to the dual-amp algorithm?
Can it not also reflect bursts detected from the bycycle-burst detection algorithm?

tutorials/plot_2_bycycle_algorithm.py
```diff
@@ -133,10 +134,10 @@
 # 3. Compute features of each cycle
 # ---------------------------------
 # After these 4 points of each cycle are localized, we compute some simple statistics for each
-# cycle. The main cycle-by-cycle function,  compute_features(), returns a table (pandas.DataFrame)
-# in which each entry is a cycle and each column is a property of that cycle (see table below).
-# There are columns to indicate where in the signal the cycle is located, but the four main features
-# are:
+# cycle. The main cycle-by-cycle function, :func:`~.compute_features`, returns a two dataframes:
```
TomDonoghue, line 137: `returns a two dataframes` -> drop the 'a'

TomDonoghue:
I did a quick review here (only looking at files from the docsite), and left some small comments. 

I agree that there might be some new tutorials, but that that is a separate topic. I think once the updates are merged in, this can go in to bring the existing documentation up to date, and then we can work on extensions from there.

TomDonoghue:
One thing that could be added to this PR (or done in another one). Right now, the tutorials load some pre-computed simulated data. These files are in `tutorials/data`, and the code that originally computed them is in `tutorials/data/sim_code.py` (though this code is for an old version of the sims). 

I think we don't need to load precomputed sim data, rather, we can simulated the needed data directly in the tutorials. Can you update the files to simulate their own data? You can follow the sims in `sim_code.py` to recreate new sims that are approximately the same as the old ones. They don't have to be exactly the same - as long as they seem good for demonstration / to make the point. (If your not sure of the current equivalent of some of the old sim code, just ask me - but from a quick glance, I think all the sims can be updated to use current NDSP sim funcs).

doc/glossary.rst
ryanhammonds: I think the definition for `Peak (time_peak)` is correct, it's the time between adjacent zero-crossings. It is defined in the docstring as the `time between rise and decay zero-crossing`, and confirmed with:
```
>>> start = df_samples.iloc[0]['sample_zerox_rise']
>>> end = df_samples.iloc[0]['sample_zerox_decay']
>>> len(sig[start:end]*fs) == df_features.iloc[0]['time_peak']
True
```
The definition you gave, I think, is for this (which is a different entry in the glossary):
```
Voltage Peak (volt_peak)
        The voltage at the current peak.
```

I can see the confusion here though. `time_peak` sounds like it should refer to the time at which the peak occurs, which is instead stored in `df_samples` as `sample_peak` (and must be multiplied by fs to give the actual time the peak occurs). Maybe it `time_peak` should be redefined as `duration_peak`?

ryanhammonds: Also, I should probably also include a glossary section for the columns defined in `df_samples`.

doc/glossary.rst
ryanhammonds: Yes, burst fraction is only defined using the dual-amp algorithm, and not the bycycle alogorithm. The bycycle burst detection algorithm returns one value (i.e. amp_fraction, amp_consistency, period_consistency, monotonicity) per cycle. The dual-amp algorithm instead returns one value per sample in a cycle (i.e if a cycle has three samples / timepoints: `np.array([True, True, False])`, the burst fraction would be 0.66.

ryanhammonds:
The latest four commits addressed some of you review comments:

- Fixed typos in the glossary (6a88d08).
- Added a glossary section for `df_samples` under the `Cyclepoints` header (6a88d08).
- Replace the dependence on sim_code.py / np.load in the tutorial with equivalent ndsp functions (ac7c4d8)
- Rebased to drop the first `unmerged prs` commit.
- Examples updated to use correct kwargs (6000db0)
- Removed loading sim_code in cyclepoint tests and replaced with ndsp sim fixture (f4497f2).

doc/glossary.rst
TomDonoghue: Ahh, I see - the difference is in terms of being defined in terms of samples vs. cycles, which I hadn't tracked properly.

doc/glossary.rst
TomDonoghue: Ohh, I see. Yeh, I had mis-interpreted this term (as something equivalent to "sample-peak"). 

I don't think the name is totally ideal. But this PR is _docs_ so, let's not start poking at names here - let's keep this name in mind, and do a sweep of API & names on the release candidate.

TomDonoghue (approved):

TomDonoghue:
Okay - I've had a look through the updates, and responded to a couple open comments on this. 

Everything looks good to me, this is ready to merge whenever you are ready @ryanhammonds !

PR #73, opened 2020-10-01
[ENH] Recompute is_burst for cycles on the edges of bursts

ryanhammonds:
This adds a utility function allowing cycles on the edges of bursts (i.e. non-burst cycles immediately to the left and right of a burst) to be recomputed using different thresholds. This address an issue noted by @sydney-smith in #59 for mislabeled cycles. In this example, the amplitude consistency threshold is lowered for edge cycles in the second plot:

```
# Load data
data = np.load('./sample_signal.npz', allow_pickle=True)
sig = data['sig']
fs = int(data['fs'])
f_range = tuple(data['f_range'])

# Compute features using a first-pass set of thresholds
threshold_kwargs = {'amp_fraction_threshold': 0.6, 'amp_consistency_threshold': 0.4,
                   'period_consistency_threshold': 0.4, 'monotonicity_threshold': 0.6, 'min_n_cycles': 2}
df_features, df_samples = compute_features(sig, fs, f_range, threshold_kwargs=threshold_kwargs)

# Lower the amp_consistency_threshold
threshold_kwargs_edges = threshold_kwargs.copy()
threshold_kwargs_edges['amp_consistency_threshold'] = 0

df_features_recompute = recompute_edges(df_features, threshold_kwargs_edges)

plot_burst_detect_summary(df_features, df_samples, sig, fs, threshold_kwargs, plot_only_result=True)
plot_burst_detect_summary(df_features_recompute, df_samples, sig, fs, threshold_kwargs, plot_only_result=True)
```
edge_fix


bycycle/burst/utils.py
```diff
@@ -49,3 +49,73 @@ def check_min_burst_cycles(is_burst, min_n_cycles=3):
...
+    """
+
+    from bycycle.burst import detect_bursts_cycles
+
+    # Prevent overwriting the orignal dataframe
+    df_features_edges = df_features.copy()
+
+    # Identify all cycles where is_burst changes on the following cycle
+    #   Use copy to keep dataframe columns unlinked
+    is_burst = df_features_edges.copy()['is_burst'].values
```
TomDonoghue, line 98: This is relatively minor - but this line computes a copy on the whole dataframe, can it not instead compute a copy of the extracted column, either as something like:
`df_features_edges['is_burst'].copy().values`
or
`from copy import deepcopy`
`deepcopy(df_features_edges.['is_burst'].values`)

These df's could presumably be quite large, and we already compute a copy of the whole df above (line 94). Copying the whole df twice (storing 3 copies) seems like probably more than we need.

bycycle/burst/utils.py
```diff
@@ -49,3 +49,73 @@ def check_min_burst_cycles(is_burst, min_n_cycles=3):
...
+    ...                                                      'sim_powerlaw': {'exp': 2}})
+    >>> threshold_kwargs = {'amp_fraction_threshold': 0., 'amp_consistency_threshold': .5,
+    ...                     'period_consistency_threshold': .5, 'monotonicity_threshold': .4,
+    ...                     'min_n_cycles': 3}
+    >>> df_features, _ = compute_features(sig, fs=1000, f_range=(8, 12), threshold_kwargs=threshold_kwargs)
+    >>> threshold_kwargs['amp_consistency_threshold'] = 0
+    >>> df_features_edges = recompute_edges(df_features, threshold_kwargs)
+    """
+
+    from bycycle.burst import detect_bursts_cycles
```
TomDonoghue, line 91: i presume this is imported within the function to avoid a circular import? Perhaps worth adding a quick comment noting that's why.

TomDonoghue:
As we chatted about, I'm a little curious to get some user feedback, and guidelines, on this. It seems a useful utility, so I think we should add it in, and ask people to explore it a bit from the release candidate, and evaluate from there (in terms of recommending guidelines for f & how to use this).  

In terms of code - all looks good! I made a couple small comments / possible suggestions! Please feel free to address them however you see fit, and then merge this in (I don't need to re-review).

bycycle/burst/utils.py
ryanhammonds: Agreed and thanks for the suggestion! A single deepcopy of just the values (as in your example) was all that was need, rather than 3 copy's.

PR #74, opened 2020-10-16
[WIP/DOC] Multi-version docsite

ryanhammonds:
This allows multiple versions of documentation to be hosted on the gh-pages sites. Currently, `make html-multi` may be ran to build the local version. I'll need to implement a `make install-multi` so it will be pushed to the gh-pages branch (this may take some trial and error). Different versions may be navigated via a dropdown menu:

multiverions_navbar

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -6,6 +6,8 @@ SPHINXBUILD   = sphinx-build
 SPHINXPROJ    = bycycle
 SOURCEDIR     = .
 BUILDDIR      = _build
+STABLE 		  = v0.1.3
+DEVELOPMENT   = v1.0.0rc
 
 # Put it first so that "make" without argument is like "make help".
 help:
@@ -52,32 +54,38 @@ install:
 	git commit -a -m 'Make install' && \
 	git push
 
-
 # Build multiple versions
 html-multi:
-	# Ensure each tag is a branch prior to running, example:
-	#   git branch v0.1.3 0.1.3
-
 	# Clean out existing build
 	make clean
-
-	# Setup html directory
-	mkdir -p _build/html_tmp
-
-	# Add versions to build here (must start with 'v')
-	for v in v0.1.3 v1.0.0rc; do \
-		mkdir -p _build/html_tmp/$$v; \
+	# Ensure navbar template exists
+	orig_branch=`git branch --show-current`
+	git checkout $(STABLE)
+	if [ ! -f _templates/navbar.html ]; then \
+		git checkout $(DEVELOPMENT) _templates/navbar.html -3; \
+		git commit -m "navbar updated"; \
+	fi
+	# Create the stable site
+	make html
+	# Cleanup stable build
+	rm -rf _build/doctrees
+	rm -rf auto_examples
+	rm -rf auto_tutorials
+	rm -rf generated
+	mv _build/* _build/html_stable
+	# Add versions to build here
+	for v in $(DEVELOPMENT); do \
 		git checkout $$v; \
 		make html; \
+		mkdir -p _build/html_stable/$$v; \
+		mv _build/html/* _build/html_stable/$$v; \
 		rm -rf auto_examples; \
 		rm -rf auto_tutorials; \
 		rm -rf generated; \
 		rm -rf _build/doctrees; \
-		mv _build/html/* _build/html_tmp/$$v; \
 		rm -rf _build/html; \
 	done
-
-	# Finalize the tmp directory
-	mv _build/html_tmp _build/html
-
-
+	# Move stable back to root reference
+	mv _build/html_stable _build/html
+	mkdir _build/html/$(STABLE)
+	cp _build/html/index.html _build/html/$(STABLE)
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -6,8 +6,9 @@ SPHINXBUILD   = sphinx-build
 SPHINXPROJ    = bycycle
 SOURCEDIR     = .
 BUILDDIR      = _build
-STABLE        = v0.1.3
-DEVELOPMENT   = v1.0.0rc
+STABLE        = 0.1.3
+DEVELOPMENT   = 1.0.0rc1
+RECENT		  = ""
 
 # Put it first so that "make" without argument is like "make help".
 help:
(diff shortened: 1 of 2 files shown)


TomDonoghue:
Hey @ryanhammonds - first off, this is awesome: super cool that you solved the multi-version docsite problem!

In terms of PR size & reviewing, I don't think we want a mega-doc-site PR. I think the site management updates can be considered one thing, and be merged in, and any new tutorials can in general each be their own new PR. 

Just to clarify, the multi-version docsites can be checked out as tags, meaning the match releases, right? That seems really nice. 

For this PR - is it ready for review / is there anything I should look at, in particular in terms of site management? If the multi-site aspect is stable, we can perhaps merge this in, and edit content from there. 

Also - this seems to be an approach we will likely apply across our modules (right?). It also seems like the kind of thing we want to document a) how the multi-site build works, and b) how maintainers should build the site (and any dependencies, such as a clear release tag approach). This is perhaps something that should go into some kind of "meta-module documentation" somewhere.

ryanhammonds:
Hi @TomDonoghue, thanks! I moved commits related to the new 2d/3d example to #79 to keep this PR focused on the multi version site.

> Just to clarify, the multi-version docsites can be checked out as tags, meaning the match releases, right? That seems really nice

Yep, tags will match release versions, assuming the tag release was used to build the distribution before uploading to pypi.

> For this PR - is it ready for review / is there anything I should look at, in particular in terms of site management?

I think it's ready for review now. I thought about pushing updates to build all versions of bycycle (like MNE does) but don't think that's needed. I think we should only have docs for the latest stable version and the development version. What are your thoughts on this?

Anyways, I think being able to build the site from this PR is the main thing for review and maybe just a quick sanity check. 

> It also seems like the kind of thing we want to document

Building the site from this branch requires several steps that I may have not described well in the post. Maybe I should but together some kind of documentation like you mentioned before reviewing.

TomDonoghue:
Quick thing - to the last point - yeh @ryanhammonds: could write out a "how to run the docsite" documentation piece? I'm not 100% where to put it (you could drop it in a comment for now), but then at least we have it. I'll review properly / fully after that!

ryanhammonds:
Multi-Version Docsite
----------------------------

1) Fetch tags, corresponding to pypi releases:
   ```
   git fetch upstream --tags
   ```
2) Create local branches from tags:
   ```
   git branch v0.1.3 tags/0.1.3
   git branch v1.0.0rc tags/1.0.0rc1
   ```

   The names of these branches should match the Makefile:
   ```
   STABLE        = v0.1.3
   DEVELOPMENT   = v1.0.0rc
   ```

3) Checkout the most recent branch:
   ```
   git checkout v1.0.0rc
   ```

4) Ensure these branches have been merged:
   ```
   git remote add ryan https://github.com/ryanhammonds/bycycle.git
   git fetch ryan
   git merge ryan/parallel # creates parallel doc example
   git merge ryan/doc      # contains multi-site and misc doc commits
   ```

   Or, if the above branches / PRs have been merged into master:
   ```
   git fetch upstream
   git merge upstream/master
   ```
   Note: This can be skipped once these branches are merged into a 1.0.0 release.

5) Navigate to docs and build from the latest dev version (i.e. v1.0.0rc):
   ```
   cd docs
   make html-multi
   ```
  Html files will be made in docs/_build.

7) Push to the gh pages branch and update site:
   ```
   make install-multi
   ```

@TomDonoghue Here are instructions on how to install.

TomDonoghue:
So, I'm trying to build the docs like this, and a couple quirks:
- the information in the original comment is outdated, and I think conflicts with later points. Please (in general) edit comments to make things consistent
- Part of the issue is perhaps that I started by trying to work from a single-branch clone of the branch with the doc update. It'll perhaps be easier to run this when things are merged. I have some thoughts / comments here, and once we've checked through these, I'll do a retake of running the build. 

Points about the instructions (in previous comment):
- the first command, `git fetch upstream --tags` doesn't run. I think you've defined 'upstream' for this to work. As far as I can tell, `git fetch origin --tags` or `git fetch --all --tags` does work (at least, in so far as enabling the next steps)
- I don't really understand 4. I roughly figured out that this is to add remote branches*, but it's not super well described that this is the case (which are 'these branches'). Also, in general, I presume this wouldn't be part of standard doc-building, right?
- I am unclear why most of these steps (say, [1, 2, 5] at least) can't be automated as part of `make-html`, or similar. Once the tags are set in the Makefile, these steps are plug and play, right?
- I'm not sure I quite understand which branch the docs should be built from. This should be more clearly specified in the instructions. I'm guessing this sets which version is 'default', right? So one should always build from the STABLE branch?
- in step 5, `doc` not `docs`

*also: you have a lot of branches on `ryanhammonds/bycycle`, which are not really very navigate-able by anyone else. For feature planned extensions / additions, I suggest using branches on the repo (ex. `bycycle-tools/bycycles/tree/multidoc` organizations. This is more transparent in terms of what things are being worked on, and easier for other people to navigate, as needed (for example, checking branches to be added to docs, etc). 

ToDo: In my mind, the previous comment describes a "how to mutli-docsite" that we should make sure is really solid, and then we can use that as general documentation for this approach, and propagate through the tools. 
@ryanhammonds - can you check through this comment, and update the instructions in prior comment, perhaps flesh it out a bit, and if possible, add more of the build steps into the Makefile directly?

Also - open question:
- This, right, builds 'STABLE' and 'DEVELOPMENT' branch. How easy would it be to add 'LAST' (name could change - but meaning the most recent prior version, before stable). I'm thinking having the prior version might be useful (for example: if one is in the middle of a project, didn't update yet, needs to check docs for prior version). Like, when we release 1.0, 1.0 will be stable, there will be no current dev version for a bit, but it might be useful to keep up for a bit. What do you think?

ryanhammonds:
Yeah, sorry! I should've noted that upstream points to bycycle-tools/bycycle. Upstream is a [convention](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/configuring-a-remote-for-a-fork) that points to the original repo (bycycle-tools/bycycle) whereas origin points to someone's own fork (ryanhammonds/bycycle). It sounds like you don't use forks though so in your case, upstream would be origin instead.

Step 4 merges this PR and #79 into the 1.0.0rc release to include the corresponding changes when building the site. Since you don't use forks, I'll start making branches on the org repos like you suggested.

The latest commit automates the entire process so now only instruction is to run `make html-multi`. I also add `RECENT` in addition to `STABLE` and `DEVELOPMENT` variables. Multiple other versions can be specified with the recent variable (space delimited). Note that 0.1.2 and older won't work because `make html` wasn't defined in doc/Makefile.

TomDonoghue:
@ryanhammonds - my apologies, I apparently totally missed your follow up message here!

Yeh, that makes sense. I tend to use branches on our modules (and forks for 3rd party), but I'm clearly not totally up to date on the conventions here! Thanks for the links!

Thanks for consolidating all the build instructions / processes into the Makefile. It all looks really smooth now - mostly docs are built same as before, and then there's just a minor update if changing which versions are hosted, which all makes sense. Nice and clean!

ryanhammonds:
@TomDonoghue No worries, thanks for the feedback! And sorry for jumping the gun on merging this in. I needed to push one more commit to cleanup the bit that pulls from my branches, since they are now merged into master. I will push a quick cleanup up to #80. Then I can make similar PRs in ndsp/fooof.

PR #75, opened 2020-10-27
[ENH] Single dataframe

ryanhammonds:
After discussing with @nschawor, I have a list of updates to make before releasing 1.0.0. This PR so far addresses returning a single df instead of two. `return_samples` kwarg now adds/removes columns, rather than returning to separate dataframes.

### DIFF
--- a/bycycle/features/burst.py
+++ b/bycycle/features/burst.py
@@ -133,8 +133,7 @@ def compute_amp_consistency(df_shape_features):
     Parameters
     ----------
     df_shape_features : pandas.DataFrame
-        Shape features for each cycle, determined using :func:`~.compute_shape_features` with
-        ``return_samples=True``.
+        Shape features for each cycle, determined using :func:`~.compute_shape_features`.
     df_samples : pandas.DataFrame
         Indices of cyclepoints returned from :func:`~.compute_cyclepoints`.
 
@@ -142,10 +141,6 @@ def compute_amp_consistency(df_shape_features):
     -------
     amp_consist : 1d array
         The amplitude consistency of each cycle.
-
-    Notes
-    -----
-    This function requires the input dataframe to contain sample indices of cyclepoints.
 
     Examples
     --------
@@ -243,7 +238,7 @@ def compute_monotonicity(df_samples, sig):
 
     Parameters
     ----------
-    df_samples: pandas.DataFrame
+    df_samples : pandas.DataFrame
         Indices of cyclepoints returned from :func:`~.compute_cyclepoints`.
     sig : 1d array
         Time series.
--- a/bycycle/features/burst.py
+++ b/bycycle/features/burst.py
@@ -9,16 +9,13 @@ from bycycle.burst import detect_bursts_dual_threshold
 ###################################################################################################
 ###################################################################################################
 
-def compute_burst_features(df_shape_features, df_samples, sig,
-                           burst_method='cycles', burst_kwargs=None):
+def compute_burst_features(df_shape_features, sig, burst_method='cycles', burst_kwargs=None):
     """Compute burst features for each cycle.
 
     Parameters
     ----------
     df_shape_features : pandas.DataFrame
         Shape parameters for each cycle, determined using :func:`~.compute_shape_features`.
-    df_samples : pandas.DataFrame
-        Indices of cyclepoints returned from :func:`~.compute_cyclepoints`.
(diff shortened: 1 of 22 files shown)


TomDonoghue:
These kind of comments should probably go into an issue, for any discussion, and then open PRs to respond (cus this isn't really the right place for comments).

- what are the pros / cons to 1 vs 2 DFs. I think at some point someone (Erik?) requested the idea of two, which might have benefits?

- what does "add self imports to all docstrings so they can be copy/pasted to run" mean? The doctests do run right, considering that we test this with doctest?

- for plotting, we want to be careful adding `plt.show` everywhere, I think. It can make for weird interactions when the user tries to use it outside the function, and also if functions get called repeatedly (for example, when we build plots with multiple plot calls). Some modules have a `show=True/False` and then call plt.show accordingly, which is a possible option. In the end it comes down to what modes we support, and we don't want to make other modes worse to use.

ryanhammonds:
@TomDonoghue I'll move these points to an issue and reference this PR.

> what are the pros / cons to 1 vs 2 DFs. I think at some point someone (Erik?) requested the idea of two, which might have benefits?

It makes thing a little simpler I suppose. The dataframe isn't too large. I wish there was some kind of sub-dataframe we could have with pandas that is shown but can still be accesses.

> what does "add self imports to all docstrings so they can be copy/pasted to run" mean? The doctests do run right, considering that we test this with doctest?
If you copy and paste the docstring examples they won't run because the function one copies from isn't imported.

> for plotting, we want to be careful adding plt.show everywhere, I think.
Yeah, plt.show will cause the plot to show twice in a notebook. A kwarg may be helpful for those you perfer ipython.

ryanhammonds:
General discussion moved to #76.

bycycle/features/burst.py
```diff
@@ -123,20 +120,21 @@ def compute_amp_fraction(df_shape_features):
...
-def compute_amp_consistency(df_shape_features, df_samples):
+def compute_amp_consistency(df_shape_features):
     """Compute amplitude consistency for each cycle.
 
     Parameters
     ----------
     df_shape_features : pandas.DataFrame
-        Shape features for each cycle, determined using :func:`~.compute_shape_features`.
+        Shape features for each cycle, determined using :func:`~.compute_shape_features` with
+        ``return_samples=True``.
```
ryanhammonds, line 137: Drop the `return_samples` bit, that has been removed from `compute_shape_features` and is only available as kwarg in `compute_features`.

bycycle/features/burst.py
```diff
@@ -145,6 +143,10 @@ def compute_amp_consistency(df_shape_features, df_samples):
     amp_consist : 1d array
         The amplitude consistency of each cycle.
 
+    Notes
+    -----
+    This function requires the input dataframe to contain sample indices of cyclepoints.
```
ryanhammonds, line 148: Same as above, remove this note.

bycycle/features/burst.py
```diff
@@ -241,7 +243,7 @@ def compute_monotonicity(df_samples, sig):
 
     Parameters
     ----------
-    df_samples : pandas.DataFrame
+    df_samples: pandas.DataFrame
```
ryanhammonds, line 246: space before colon

bycycle/plts/burst.py
```diff
@@ -71,9 +70,8 @@ def plot_burst_detect_summary(df_features, df_samples, sig, fs, threshold_kwargs
     >>> sig = sim_bursty_oscillation(10, fs, freq=10)
     >>> threshold_kwargs = {'amp_fraction_threshold': 0., 'amp_consistency_threshold': .5,
     ...                     'period_consistency_threshold': .5, 'monotonicity_threshold': .8}
-    >>> df_features, df_samples = compute_features(sig, fs, f_range=(8, 12),
-    ...                                            threshold_kwargs=threshold_kwargs)
-    >>> plot_burst_detect_summary(df_features, df_samples, sig, fs, threshold_kwargs)
+    >>> df_features= compute_features(sig, fs, f_range=(8, 12), threshold_kwargs=threshold_kwargs)
```
ryanhammonds, line 73: space for equals sign

ryanhammonds:
@TomDonoghue I did a quick self-review and fixed a few typos. I also added a util helper func to separate df_samples from df_features. This PR should be ready to merge but I'll hold off for now.

Issue #76, opened 2020-10-27
[ENH] 1.0.0rc feedback


ryanhammonds:
I met with @nschawor, who proposed:

- [x] Return a single df instead of two. return_samples kwarg now adds/removes columns, rather than returning to separate dataframes (#75). Edit: A utility func has been added allow df_samples to be separate from df_features if desired.
- [x] Increased flexibility for compute_features_2d and compute_features_3d (global thresholds / edges).
- [x] Update and add new examples/tutorials.
- [x] ~~Rename compute_burst_features~~
  - this doesn't actually do burst detection (that's done in burst.detect_bursts_amp) so I don't think this func needs renaming.
- [x] Plotting in jupyter vs ipython
  - plots need plt.show if not in jupyter
- [x] Add self imports to all docstrings so they can be copy/pasted to run.
  - This is something that needs to be updated across all repos (ndsp, fooof, etc)

@TomDonoghue brought up these points in #75:

> - what are the pros / cons to 1 vs 2 DFs. I think at some point someone (Erik?) requested the idea of two, which might have benefits?

It makes thing a single df little simpler I suppose. The dataframe isn't too large. I wish there was some kind of sub-dataframe we could have with pandas that is ~~shown~~ hidden but can still be accessed. Maybe @nschawor could chime in on this.

> - what does "add self imports to all docstrings so they can be copy/pasted to run" mean? The doctests do run right, considering that we test this with doctest?

If one copies/pastes the docstring examples they won't run because the function the docstring is from isn't imported.

> - for plotting, we want to be careful adding plt.show everywhere, I think. It can make for weird interactions when the user tries to use it outside the function, and also if functions get called repeatedly (for example, when we build plots with multiple plot calls). Some modules have a show=True/False and then call plt.show accordingly, which is a possible option. In the end it comes down to what modes we support, and we don't want to make other modes worse to use.

I agree that adding plt.show could be problematic since it would display the plot twice in a notebook (I think). A kwarg may be helpful for those you prefer ipython (or what @nschawor prefers to use, I can't remember what it was).

Anyone else's feedback from exploring 1.0.0rc is welcome here!

TomDonoghue:
Small / quick follow ups:

***dataframes***
[...]

PR #77, opened 2020-10-27
[ENH] 2d/3d Array Improvement

ryanhammonds:
**2d Array**
This addresses a point in #76 and is built on top of #75 (i.e. starting from deabc11). It allows for 2d arrays to be flatten prior to computing features using `axis = None`. The resulting dataframe is split into a list of dfs corresponding to indices in the original 2d signal. This is to address edge artifacts and features miscalculations from ignoring the influence of other signals.

For example flattening 2d to 1d (`axis=None`), or treating each signal separately (`axis=0`) will result in different burst fractions (`compute_burst_fraction`):
```
df_shape_features['volt_amp'].rank() / len(df_shape_features)
```

 - ``axis=1`` : Iterates over each row/signal in an array independently (i.e. for each channel
   in (n_channels, n_timepoints)).
 - ``axis=None`` : Flattens rows/signals prior to computing features (i.e. across flatten epochs
   in (n_epochs, n_timepoints)).

Another note, is that when  `axis = None`, the 2d to 1d array conversion prevents parallelization.

**3d Array**
An axis kwarg may be passed in `compute_features_3d`:

 - ``axis=0`` : Iterates over 2D slices along the zeroth dimension, (i.e. for each channel in
   (n_channels, n_epochs, n_timepoints)).
 - ``axis=1`` : Iterates over 2D slices along the first dimension (i.e. across flatten epochs in
   (n_epochs, n_timepoints)).
 - ``axis=(0, 1)`` : Iterates over 1D slices along the zeroth and first dimension (i.e across
   each signal independently in (n_participants, n_channels, n_timepoints)).

### DIFF
--- a/bycycle/features/burst.py
+++ b/bycycle/features/burst.py
@@ -133,8 +133,7 @@ def compute_amp_consistency(df_shape_features):
     Parameters
     ----------
     df_shape_features : pandas.DataFrame
-        Shape features for each cycle, determined using :func:`~.compute_shape_features` with
-        ``return_samples=True``.
+        Shape features for each cycle, determined using :func:`~.compute_shape_features`.
     df_samples : pandas.DataFrame
         Indices of cyclepoints returned from :func:`~.compute_cyclepoints`.
 
@@ -142,10 +141,6 @@ def compute_amp_consistency(df_shape_features):
     -------
     amp_consist : 1d array
         The amplitude consistency of each cycle.
-
-    Notes
-    -----
-    This function requires the input dataframe to contain sample indices of cyclepoints.
 
     Examples
     --------
@@ -243,7 +238,7 @@ def compute_monotonicity(df_samples, sig):
 
     Parameters
     ----------
-    df_samples: pandas.DataFrame
+    df_samples : pandas.DataFrame
         Indices of cyclepoints returned from :func:`~.compute_cyclepoints`.
     sig : 1d array
         Time series.
--- a/bycycle/features/burst.py
+++ b/bycycle/features/burst.py
@@ -9,16 +9,13 @@ from bycycle.burst import detect_bursts_dual_threshold
 ###################################################################################################
 ###################################################################################################
 
-def compute_burst_features(df_shape_features, df_samples, sig,
-                           burst_method='cycles', burst_kwargs=None):
+def compute_burst_features(df_shape_features, sig, burst_method='cycles', burst_kwargs=None):
     """Compute burst features for each cycle.
 
     Parameters
     ----------
     df_shape_features : pandas.DataFrame
         Shape parameters for each cycle, determined using :func:`~.compute_shape_features`.
-    df_samples : pandas.DataFrame
-        Indices of cyclepoints returned from :func:`~.compute_cyclepoints`.
     sig : 1d array
         Voltage time series used for determining monotonicity.
     burst_method : string, optional, default: 'cycles'
@@ -63,8 +60,8 @@ def compute_burst_features(df_shape_features, df_samples, sig,
     >>> from neurodsp.sim import sim_bursty_oscillation
     >>> fs  = 500
     >>> sig = sim_bursty_oscillation(10, fs, 10)
-    >>> df_shapes, df_samples = compute_shape_features(sig, fs, f_range=(8, 12))
-    >>> df_burst = compute_burst_features(df_shapes, df_samples, sig, burst_method='amp',
+    >>> df_shapes = compute_shape_features(sig, fs, f_range=(8, 12))
+    >>> df_burst = compute_burst_features(df_shapes, sig, burst_method='amp',
     ...                                   burst_kwargs={'fs': fs, 'f_range': (8, 12)})
     """
 
@@ -78,10 +75,10 @@ def compute_burst_features(df_shape_features, df_samples, sig,
         df_burst_features['amp_fraction'] = compute_amp_fraction(df_shape_features)
 
         df_burst_features['amp_consistency'] = \
-                compute_amp_consistency(df_shape_features, df_samples)
+                compute_amp_consistency(df_shape_features)
 
         df_burst_features['period_consistency'] = compute_period_consistency(df_shape_features)
-        df_burst_features['monotonicity'] = compute_monotonicity(df_samples, sig)
+        df_burst_features['monotonicity'] = compute_monotonicity(df_shape_features, sig)
 
     # Use dual threshold burst detection
     elif burst_method == 'amp':
@@ -94,7 +91,7 @@ def compute_burst_features(df_shape_features, df_samples, sig,
                              "when 'burst_method' is 'amp'.")
 
         df_burst_features['burst_fraction'] = \
-            compute_burst_fraction(df_samples, sig, fs, f_range, **burst_kwargs)
+            compute_burst_fraction(df_shape_features, sig, fs, f_range, **burst_kwargs)
 
     else:
         raise ValueError("Unrecognized 'burst_method'.")
@@ -123,20 +120,21 @@ def compute_amp_fraction(df_shape_features):
     >>> from neurodsp.sim import sim_bursty_oscillation
     >>> fs = 500
     >>> sig = sim_bursty_oscillation(10, fs, freq=10)
-    >>> df_shapes, df_samples = compute_shape_features(sig, fs, (8, 12))
+    >>> df_shapes = compute_shape_features(sig, fs, (8, 12))
     >>> amp_fraction = compute_amp_fraction(df_shapes)
     """
 
     return df_shape_features['volt_amp'].rank() / len(df_shape_features)
 
 
-def compute_amp_consistency(df_shape_features, df_samples):
+def compute_amp_consistency(df_shape_features):
     """Compute amplitude consistency for each cycle.
 
     Parameters
     ----------
     df_shape_features : pandas.DataFrame
-        Shape features for each cycle, determined using :func:`~.compute_shape_features`.
+        Shape features for each cycle, determined using :func:`~.compute_shape_features` with
+        ``return_samples=True``.
     df_samples : pandas.DataFrame
         Indices of cyclepoints returned from :func:`~.compute_cyclepoints`.
 
@@ -144,6 +142,10 @@ def compute_amp_consistency(df_shape_features, df_samples):
     -------
     amp_consist : 1d array
         The amplitude consistency of each cycle.
+
+    Notes
+    -----
+    This function requires the input dataframe to contain sample indices of cyclepoints.
 
     Examples
     --------
@@ -153,8 +155,8 @@ def compute_amp_consistency(df_shape_features, df_samples):
     >>> from neurodsp.sim import sim_bursty_oscillation
     >>> fs = 500
     >>> sig = sim_bursty_oscillation(10, fs, freq=10)
-    >>> df_shapes, df_samples = compute_shape_features(sig, fs, f_range=(8, 12))
-    >>> amp_consistency = compute_amp_consistency(df_shapes, df_samples)
+    >>> df_shapes = compute_shape_features(sig, fs, f_range=(8, 12))
+    >>> amp_consistency = compute_amp_consistency(df_shapes)
     """
 
     # Compute amplitude consistency
@@ -170,7 +172,7 @@ def compute_amp_consistency(df_shape_features, df_samples):
 
             consist_current = np.min([rises[cyc], decays[cyc]]) / np.max([rises[cyc], decays[cyc]])
 
-            if 'sample_peak' in df_samples.columns:
+            if 'sample_peak' in df_shape_features.columns:
 
                 consist_last = np.min([rises[cyc], decays[cyc-1]]) / \
                     np.max([rises[cyc], decays[cyc-1]])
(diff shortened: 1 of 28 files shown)


ryanhammonds:
This should be ready for a review. I've made a tutorial for how/when to use these functions in #74 and it has been pushed to the docsite [here](https://bycycle-tools.github.io/bycycle/v1.0.0rc/auto_examples/plot_2_multidim_sigs.html#sphx-glr-auto-examples-plot-2-multidim-sigs-py). Hopefully the tutorial makes sense! I would start with the tutorial before reviewing the code.

ryanhammonds:
Maybe add 2D/3D plotting funcs that stack 1D plots? MNE has nice functionality that lets you plot epochs vertically with interactive annotation / zooming features too - something I've looked into adding to ndspflow.

TomDonoghue:
Ahhh, sorry @ryanhammonds - I seemed to have missed to tag to review this one! I'll get to it soon!

bycycle/features/features.py
```diff
@@ -83,13 +83,11 @@ def compute_features(sig, fs, f_range, center_extrema='peak', burst_method='cycl
...
 
-        When dual threshold burst detection is used (i.e. burst_method == 'amp'):
+        When dual threshold burst detection is used (i.e. burst_method = 'amp'):
 
         - ``burst_fraction`` : fraction of a cycle that is bursting
 
-    df_samples : pandas.DataFrame, optional, default: True
-        An optionally returned dataframe containing cyclepoints for each cycle.
-        Columns (listed for peak-centered cycles):
+        When cyclepoints are returned (i.e. deafault, return_samples = True)
```
TomDonoghue, line 90: typo on `deafault`

TomDonoghue: also - in comments do / should we write `return_samples = True` or `return_samples=True`. 
It looks a little weird to me to write this differently in comment than we would in code.

bycycle/features/features.py
```diff
@@ -150,7 +146,8 @@ def compute_features(sig, fs, f_range, center_extrema='peak', burst_method='cycl
         raise ValueError('Invalid argument for "burst_method".'
                          'Either "cycles" or "amp" must be specified."')
 
-    if return_samples:
-        return df_features, df_samples
+    if return_samples is False:
```
TomDonoghue, line 149: could / should this be a dataframe util? Just like we have 'split_samples_df', it could be `drop_samples_df` or something. I think it would be useful to have the utility if there may be some cases to do this operation outside this function.

bycycle/features/burst.py
```diff
@@ -9,16 +9,13 @@
...
 ###################################################################################################
 
-def compute_burst_features(df_shape_features, df_samples, sig,
-                           burst_method='cycles', burst_kwargs=None):
+def compute_burst_features(df_shape_features, sig, burst_method='cycles', burst_kwargs=None):
     """Compute burst features for each cycle.
 
     Parameters
     ----------
     df_shape_features : pandas.DataFrame
```
TomDonoghue, line 17: doesn't this input now combine both shape & cyclepoint features? If so, this description seems perhaps slightly out of data. Maybe it should be called `df_features`?

bycycle/group/features.py
```diff
@@ -74,91 +91,154 @@ def compute_features_2d(sigs, fs, f_range, compute_features_kwargs=None,
...
+        sig_flat = sigs.flatten()
 
-        else:
-            df_features = list(progress_bar(mapping, progress, len(sigs)))
+        center_extrema = kwargs[0].pop('center_extrema', 'peak')
+
+        df_flat = compute_features(sig_flat, fs=fs, f_range=f_range, return_samples=True,
+                                   center_extrema=center_extrema, **kwargs[0])
+
+        # Reshape the dataframe into original sigs shape
```
TomDonoghue, line 137: so, this feels a bit like it's an independent, and potentially useful task: to 'epoch' a dataframe of measured values. 

Could this be a sub-function? As well as increasing modularity, I feel like it could be useful be itself. For example, if I've already computed features on a long continuous signal, I can imagine wanting to split it up into time sections. 

Is there a relatively easy way to turn this code section into something like:
`epoch_df(df, epoch_lengths)`
which would take a single df, and split it up into a list of dfs, each representing `epoch_length` samples?
This function could be used here, and offered as a general utility, I think?

bycycle/group/features.py
```diff
@@ -189,81 +269,90 @@ def compute_features_3d(sigs, fs, f_range, compute_features_kwargs=None,
...
+    check_kwargs_shape(sigs, kwargs, axis)
+    kwargs = list(kwargs.flatten()) if isinstance(kwargs, np.ndarray) else [kwargs]
 
-    # Ensure compute_features corresponds to sigs
-    if isinstance(kwargs, np.ndarray):
+    # Compute features
+    df_features = np.zeros((np.shape(sigs)[0], np.shape(sigs)[1])).tolist()
 
-        if kwargs.ndim == 1 and np.shape(kwargs)[0] != np.shape(sigs)[0]:
+    if axis == 0:
```
TomDonoghue, line 285: The process for axis 0 or 1 is effectively the same, with just a bit of data manipulation depending which actual axis. 

This can therefore potentially be condensed, I think:

```
if axis in [0, 1]:
    sigs = np.swapaxes(sigs, 0, 1) is axis == 1 else sigs
    # then have the lines 287 to 295, that are otherwise duplicated in 300 to 308
    if axis == 1: df_features = [list(dfs) for dfs in zip(*df_features)]
```

bycycle/group/features.py
```diff
@@ -74,91 +91,154 @@ def compute_features_2d(sigs, fs, f_range, compute_features_kwargs=None,
...
+    axis : {0, 1, None}, optional, default: 1
+        Which axes to calculate features across:
+
+
+        - ``axis = 0`` : Features are computed for each signal independently
+          across the first axis, i.e. across channels in (n_channels, n_epochs, n_timepoints).
+        - ``axis = 1`` (default) : Features are computed for each signal independently
+          across the second axis, i.e. across channels in (n_epochs, n_channels, n_timepoints).
+        - ``axis = None`` : Features are computed across a flattened 1d array, this is typically
+          not recommended unless the 3d array was recored continuously.
```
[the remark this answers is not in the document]
TomDonoghue: Ohhh, wait. it's only when I finished reading the code that I realized this flattens -> 2D, and not all the way to 1D. 
in that case, this option feels equally sensible to the flatten option within `compute_features_2d`. I'm not sure either are super general, but I think they can plausibly be kept. But - this doc should be clear as to how it flattens.

bycycle/group/features.py
```diff
@@ -189,81 +269,90 @@ def compute_features_3d(sigs, fs, f_range, compute_features_kwargs=None,
...
-    min_idx = [i for i in range(0, len(df_features), dim_b)]
+def _proxy_3d(args, fs=None, f_range=None, return_samples=None):
+    """Proxy function to map kwargs and 3d sigs together."""
 
-    for lower, upper in zip(min_idx, max_idx):
-        df_reshape.append(df_features[lower:upper])
+    sigs, kwargs = args[0], args[1]
 
-    return df_reshape
+    return compute_features_2d(sigs, fs=fs, f_range=f_range, return_samples=return_samples,
```
TomDonoghue, line 357: very much a nitpick, but for consistency, I would say order of arguments here should also match the function def.

TomDonoghue (changes requested):

TomDonoghue:
okay @ryanhammonds - I did and finished the deep dive review here. I think overall it looks nice, and has a nice element of being able to distribute the core functions across dimensions, allowing for lots of flexibility, without too much code overall. 

So, general approach is good, as usual I've left a bunch of specific comments and suggestions along the way! :)

bycycle/group/features.py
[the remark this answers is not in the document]
ryanhammonds: The 3D array does eventually get flatten to 1D (within `compute_features_2d`), when `axis=None`. The 2d func is used in this case to maintain a 2D dataframe structure that can more easily be reorganized into a 3D list of dataframes. I added this option to make the function as flexible as possible for all use cases. But now that you mention it, I don't think this would ever be useful and can probably be removed.

bycycle/group/features.py
```diff
@@ -74,91 +91,154 @@ def compute_features_2d(sigs, fs, f_range, compute_features_kwargs=None,
...
+        Which axes to calculate features across:
+
+
+        - ``axis = 0`` : Features are computed for each signal independently
+          across the first axis, i.e. across channels in (n_channels, n_epochs, n_timepoints).
+        - ``axis = 1`` (default) : Features are computed for each signal independently
+          across the second axis, i.e. across channels in (n_epochs, n_channels, n_timepoints).
+        - ``axis = None`` : Features are computed across a flattened 1d array, this is typically
+          not recommended unless the 3d array was recored continuously.
+        - ``axis = (0, 1)`` : Features are computed independently for each signal, across the first
```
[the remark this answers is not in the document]
ryanhammonds: (0, 1) or tuples in general are supported by numpy. However, I realized axis=(0, 1) functions differently in np than it does here. In np , (0, 1) means to run across axis 0, take the resulting array and run across axis 1. I think it should be updated to axis=2 to match numpy.

```
In [1]: import numpy as np                                                                                                                   

In [2]: arr = np.ones((2, 3, 4))                                                                                                             

In [3]: arr                                                                                                                                  
Out[3]: 
array([[[1., 1., 1., 1.],
        [1., 1., 1., 1.],
        [1., 1., 1., 1.]],

       [[1., 1., 1., 1.],
        [1., 1., 1., 1.],
        [1., 1., 1., 1.]]])

In [4]: np.sum(arr, axis=0)                                                                                                                  
Out[4]: 
array([[2., 2., 2., 2.],
       [2., 2., 2., 2.],
       [2., 2., 2., 2.]])

In [5]: np.sum(arr, axis=1)                                                                                                                  
Out[5]: 
array([[3., 3., 3., 3.],
       [3., 3., 3., 3.]])

In [6]: np.sum(arr, axis=(0, 1))                                                                                                             
Out[6]: array([6., 6., 6., 6.])

In [7]: np.sum(arr, axis=2)                                                                                                                  
Out[7]: 
array([[4., 4., 4.],
       [4., 4., 4.]])
```

bycycle/group/features.py
```diff
@@ -189,81 +269,90 @@ def compute_features_3d(sigs, fs, f_range, compute_features_kwargs=None,
...
-        kwargs = np.array(kwargs)
-    elif kwargs is None:
-        kwargs = {}
+    check_kwargs_shape(sigs, kwargs, axis)
+    kwargs = list(kwargs.flatten()) if isinstance(kwargs, np.ndarray) else [kwargs]
 
-    # Ensure compute_features corresponds to sigs
-    if isinstance(kwargs, np.ndarray):
+    # Compute features
+    df_features = np.zeros((np.shape(sigs)[0], np.shape(sigs)[1])).tolist()
```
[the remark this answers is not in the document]
ryanhammonds: Good catch, thanks!

I renamed `df_features` to `dfs_features`.

tutorials/plot_3_bycycle_resting_state.py
```diff
@@ -116,8 +116,8 @@
 # periods of the signal that appear to be bursting. This was confirmed by looking at a few different
 # signal segments from a few subjects.
 
-plot_burst_detect_summary(df_features_list[1], df_samples_list[1], sigs[1], fs,
-                          threshold_kwargs, xlim=(0, 5), figsize=(16, 3))
+plot_burst_detect_summary(df_features_list[1], sigs[0], fs, threshold_kwargs,
```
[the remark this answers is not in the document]
ryanhammonds: It was a typo, good catch!

bycycle/utils/dataframes.py
```diff
@@ -102,19 +103,22 @@ def get_extrema_df(df):
...
+def rename_extrema_df(center_extrema, df_features, return_samples=True):
     """Rename a dataframe based on the centered extrema.
 
     Parameters
     ----------
     center_extrema : {'trough', 'peak'}
         Which extrema is centered.
-    df_samples, df_features : pandas.DataFrames
-        ByCycle dataframes to rename, given the centered extrema.
+    df_shape_features : pandas.DataFrames
```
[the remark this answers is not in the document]
ryanhammonds: Yep, thanks!

bycycle/group/features.py
ryanhammonds: I refactored this to a new func  in `utils.dataframes`.

bycycle/group/features.py
ryanhammonds: Good idea! Updated to this

ryanhammonds:
@TomDonoghue I addressed each of your comments. Thanks for the feedback!

I think the biggest change in the last few commits was changing how the axis arg works. It functions the exact same as the numpy axis arg now. I replaced `axis=(0, 1)`  with `axis=2` for 3D array. `axis=0` was also replaced with `axis=1` for 2D arrays. This was to functionally match the numpy axis arg.

This should be ready to merge, but I'm going to finish updating the tutorials before merging.

TomDonoghue:
@ryanhammonds - sounds good - I had a quick scroll, and all looks good to me!

PR #78, opened 2020-11-03
[ENH] ipython plotting

ryanhammonds:
Address a point in #76 for those who use ipython rather than jupyter. One line was added to the end of each plot function:
```
plt.show() if os.environ['_'].endswith('ipython') else None
```
When in jupyter, `os.environ['_']` will instead end with `jupyter-notebook`.

**Note**
The tutorials use ndsp plotting funcs (plot_time_series, etc) and this PR won't help in those cases. A separate PR will need to be made in ndsp for these cases.

### DIFF
--- a/bycycle/plts/burst.py
+++ b/bycycle/plts/burst.py
@@ -1,5 +1,6 @@
 """Plot burst detection parameters."""
 
+from os import environ
 from itertools import cycle
 
 import numpy as np
@@ -148,6 +149,9 @@ def plot_burst_detect_summary(df_features, sig, fs, threshold_kwargs, xlim=None,
                                     figsize=figsize, ax=axes[idx+1], xlim=xlim, xlabel=xlabel,
                                     ylabel=ylabel, color=color, interp=interp)
 
+    # Use show method if in ipython
+    plt.show() if environ['_'].endswith('ipython') else None
+
 
 @savefig
 def plot_burst_detect_param(df_features, sig, fs, burst_param, thresh,
@@ -266,3 +270,6 @@ def plot_burst_detect_param(df_features, sig, fs, burst_param, thresh,
             ax.axvspan(times[int(cyc['sample_last_' + side_e])],
                        times[int(cyc['sample_next_' + side_e])],
                        alpha=0.5, color=color, lw=0)
+
+    # Use show method if in ipython
+    plt.show() if environ['_'].endswith('ipython') else None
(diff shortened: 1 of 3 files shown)


ryanhammonds:
@TomDonoghue If this looks like a good solution to you, I can go ahead and make similar PRs for ndsp and fooof.

TomDonoghue:
I don't think this will work with our plot builds. 

Consider this set of functions I've been playing with in iPython:
```
def base_no(ax): ax.plot([1, 2], [3, 4])
def base_sh(ax): ax.plot([1, 2], [3, 4]); plt.show()     
def mult_no(ax): base_no(ax); ax.plot([1, 2], [5, 6]);
def mult_sh(ax): base_sh(ax); ax.plot([1, 2], [5, 6]); plt.show()

# you can play around with these funcs in ipython, with these:
import matplotlib.pyplot as plt
_, ax = plt.subplots()   
```

`mult_no` and `mult_sh` should give the same plot, with two lines. But when calling `mult_sh`, with a passed in axis, execution gets halted and shows the plot with only one line (from `base_sh`). It then doesn't show the final / proper plot. `mult_no` works as expected (with the user doing a final `plt.show`). 

So, based on our approach of having functions that take axis objects, and update them, this doesn't really work, as plt.show interrupts execution where in the process it is. If we were to add a `plt.show` line, it would have to be enforced to happen at the end of the entire plot built. That's actually tricky to do, with potentially multiple functions involved. It's _maybe_ more plausible with some kind of decorator, but I'm not so tempted to try and play that game either...

Overall, even trying to do (check the environment, proceed differently) feels bit hacky to me. I don't think we'll find a clean solution on this (that isn't some combination of a) annoying b) hard to do, c) full of side effects). In general, we don't want to play the game of "guess what platform our users are using, and manage the _n_ quirks it has". The cleanest solution is to delineate what we support, and how. Overall, I don't see a way to do solve this, nor a clear justification for dedicating time to it.

ryanhammonds:
Thanks for the example. Yeah, this PR is broken anytime there are multiple plot calls (I tested this on a function with a single plot call, whoops).

The only way for this to work is to use the plt.show line in a decorator to ensure it's executed once at the very end, or to use something like MNE does with `matplotlib.get_backend()` to be less hacky.

Either way, this is a bit more complicated than I thought. Since you don't think it's worth the effort, I'll close without merging.

PR #79, opened 2020-11-13
[ENH] 2D/3D Example

ryanhammonds:
This PR starts at 0bae741 everything before this is in #74. An easy way to ignore commits that are specific to #74, when reviewing, is to skip all doc/* file diffs. This PR contains an example for using 2D and 3d compute_features functions. It also contains various fixes for the api/examples/tutorials.


doc/changelog.rst
```diff
@@ -0,0 +1,84 @@
...
+
+Code Oraganization
+~~~~~~~~~~~~~~~~~~
+
+- Increased modularity.
+
+  - burst: ``burst.amp``, ``burst.cycle``, ``burst.dualthresh``, ``burst.utils``
+  - cyclepoints: ``cyclepoints.extrema``, ``cyclepoints.zerox``, ``cyclepoints.phase``
+  - features: ``features.features``, ``features.burst``, ``features.shape``, ``features.cyclepoints``
+  - The dataframe output returned from ``compute_features`` now returns:
```
TomDonoghue, line 26: this is outdated, I think?

doc/changelog.rst
```diff
@@ -0,0 +1,84 @@
...
+to reflect the new changes.
+
+Note that the main changes are in code organization, some names, and the addition of many new
+features. The algorithm itself has not changed, and results from the new version should be roughly
+equivalent to those with older versions, with a few minor exceptions.
+
+Code Oraganization
+~~~~~~~~~~~~~~~~~~
+
+- Increased modularity.
```
TomDonoghue, line 21: It's not 100% obvious what this describes (that these are new files, right?). I think this needs expanding, like:
`The module has been re-organized to increase modularity. Sub-folders and files are now organized as:`

doc/changelog.rst
```diff
@@ -0,0 +1,84 @@
...
+
+  - burst: ``burst.amp``, ``burst.cycle``, ``burst.dualthresh``, ``burst.utils``
+  - cyclepoints: ``cyclepoints.extrema``, ``cyclepoints.zerox``, ``cyclepoints.phase``
+  - features: ``features.features``, ``features.burst``, ``features.shape``, ``features.cyclepoints``
+  - The dataframe output returned from ``compute_features`` now returns:
+
+    1. ``df_features``: burst/shape cycle features.
+    2. ``df_samples``: cyclepoint locations as signal indices.
+
+- Neurodsp dependency.
```
TomDonoghue, line 31: `Add NeuroDSP dependency`

(otherwise, not clear if we're adding / removing, etc).

doc/changelog.rst
```diff
@@ -0,0 +1,84 @@
...
+
+  - `filt.py` and `sim.py` have been replaced with equivalent neurodsp functions.
+
+Naming Updates
+~~~~~~~~~~~~~~
+
+- Consistency with neurodsp and PEP8 compliance:
```
TomDonoghue, line 38: I actually don't think it's all that important to split this up by different reasons. From the users perspective, what matters is these have all changed (there is nothing actionable different about the different groups). I would probably just say:
`The following argument names have been changed throughout the module:`
(then one big list)

doc/changelog.rst
```diff
@@ -0,0 +1,84 @@
...
+  - Monotonicity computation didn't include all points in a cycle.
+  - Zero-crossing off-by-one error when a signal crossed at exactly zero.
+
+Documentation
+~~~~~~~~~~~~~
+
+The 1.X.X series comes with an updated documentation site.
+
+As well as updating the tutorials, API list, and other existing documentation, there are
+also (upcoming) new materials.
```
TomDonoghue, line 84: can drop the `(upcoming)`. These will be in when we merge to master / post as stable release.

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
+"""
+2. 2D and 3D Signals
+====================
+Bycycle supports computing features for 2D and 3D arrays of signals, using
```
TomDonoghue, line 4: The first sentence in a doc file should be a short, stand-alone description 

If you go to https://bycycle-tools.github.io/bycycle/v1.0.0rc/auto_examples/index.html, and hover over an example, you can see that these first lines define a pop-up description. Right now, these aren't set up to work well with ByCycle (compare to FOOOF or NeuroDSP). 

It should be:
- title
- blank line
- stand alone description line (to be used as pop-up)
- blank line
- start of main text example

This can be updated across all ByCycle tutorials & examples.

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
+"""
+2. 2D and 3D Signals
```
TomDonoghue, line 2: I'd vote for a more descriptive title. Maybe "Running ByCycle on Multi-Dimensional Arrays"?

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
...
+====================
+Bycycle supports computing features for 2D and 3D arrays of signals, using
+:func:`~.compute_features_2d` or :func:`~.compute_features_3d`. These signals may be organized in a
+variety of ways, such as (n_epochs, n_timepoints), (n_channels, n_timepoints), or
+(n_channels,  n_epochs, n_timepoints). The ``axis`` keyword argument may be used to handle various
+organizations of signals. Understanding the axis argument is important to ensure features are
+computed in an appropriate manor.
+
+For 2D arrays:
```
TomDonoghue, line 11: Having these axis descriptions here feels like data dump, with no clear way to use this information yet. Each respective axis description should be moved to be with respective examples below that do 2D / 3D examples.

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
...
+n_seconds = 10
+fs = 500
+
+freq = 10
+f_range = (5, 15)
+
+epochs = 10
+channels = 5
+
+####################################################################################################
```
TomDonoghue, line 56: extra section break

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
...
+#
+# Example 1. 2D Array (n_epochs, n_timepoints)
+# --------------------------------------------
+# The features for a 2d array with a (n_epochs, n_timepoints) organiation will be computed here.
+# This is an example of when using ``axis = None`` is appropriate, assuming each epoch was recorded
+# continuously. Data will be simulated to produce to alternating epoch types (fixation and task)
+# with varying rise-decay symmetries.
+
+# Simulate a timeseries
```
TomDonoghue, line 67: add section break

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
...
+f_range = (5, 15)
+
+epochs = 10
+channels = 5
+
+####################################################################################################
+
+####################################################################################################
+#
+# Example 1. 2D Array (n_epochs, n_timepoints)
```
TomDonoghue, line 60: before you introduce this specific 2D example, introduce 2D examples in general here, and add the axis info (for 2D only) about here

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
...
+# --------------------------------------------
+# The features for a 2d array with a (n_epochs, n_timepoints) organiation will be computed here.
+# This is an example of when using ``axis = None`` is appropriate, assuming each epoch was recorded
+# continuously. Data will be simulated to produce to alternating epoch types (fixation and task)
```
TomDonoghue, line 64: alternating epoch types seems like an odd organization, and not a super common one, I don't think. It also adds some extra, and slightly un-obvious code, for doing skip-by-2 a couple times. 

Why not just set up to have two 2D arrays, one for 'task', and one for 'rest'? 
This also allows for more of a "build up", as in:
- first lets imagine (simulate) a 2D array of rest trials
- we can apply bycycle like this
- let's look at results
- okay, cool, now let's compare to another condition. Here's a new 2D array of 'task'
- we can compute measures again
- here's how we compare

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
...
+    if idx % 2 == 0:
+        # Fixation epoch
+        sim_components = {'sim_powerlaw': dict(exponent=-2),
+                          'sim_bursty_oscillation': dict(freq=10, cycle='asine',
+                                                         enter_burst=0.75, rdsym=0.5)}
+    else:
+        # Task epoch
+        sim_components = {'sim_powerlaw': dict(exponent=-2),
+                          'sim_bursty_oscillation': dict(freq=10, cycle='asine',
+                                                         enter_burst=0.75, rdsym=0.75)}
```
TomDonoghue, line 81: I would make it more obvious what you're simulating as the difference (rdsym). 
Perhaps easiest to do so, explicit noting this, using variables. Like"
```
For this example, we will simulate different features for the rdsym value, and then see if we can reconstruct this different by applying ByCycle

# Define rdsym values for rest and task trials
rdsym_rest = 0.5
rdsym_task = 0.75
```

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
...
+
+    sigs[idx] = sim_combined(n_seconds, fs, components=sim_components)
+
+####################################################################################################
+
+# Compute features
+thresholds = dict(amp_fraction_threshold=0., amp_consistency_threshold=.5,
```
TomDonoghue, line 88: In a tutorial, we might want to add a note about where these setting values come from. Are they chosen as defaults?

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
...
+# Plot
+plot_feature_categorical(df_bursts, 'time_rdsym', group_by='epoch', ylabel='Rise-Decay Symmetry',
+                         xlabel=['Fixation', 'Task'])
+
+####################################################################################################
+#
+# Here we show what would happen if axis = 0 was incorrectly used when computing the features
+# for an (n_epochs, n_timeseries) array. If bursts exist at the edges of epochs, they will be
+# incorrectly labeled as non-burst cycles. This is due to the inability to compute consistency
+# features for the first and last cycle of each signal.
```
TomDonoghue, line 118: I found this point a bit confusing - it doesn't really describe _why_ this is the case, which is because there is a difference between computing when assuming continuity between epochs, and when not.

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
...
+    Number of bursts (axis = None): {n_bursts_axis_None}
+    Number of bursts (axis = 0   ): {n_bursts_axis_0}
+""".format(n_bursts_axis_None=len(df_bursts), n_bursts_axis_0=len(df_bursts_axis0)))
+
+####################################################################################################
+#
+# Example 2. 2D Array (n_copies, n_timepoints)
+# --------------------------------------------
+# Features for a 2D array with an (n_copies, n_timepoints) organization will be computed
+# independently, using axis = 0. Each signal will be indentical, demonstrating the effect
```
TomDonoghue, line 134: typo: 'indentical'

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
...
+epochs = 10
+channels = 5
+
+####################################################################################################
+
+####################################################################################################
+#
+# Example 1. 2D Array (n_epochs, n_timepoints)
+# --------------------------------------------
+# The features for a 2d array with a (n_epochs, n_timepoints) organiation will be computed here.
```
TomDonoghue, line 62: typo on 'organiation'

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
...
+print("""
+    Number of bursts (axis = None): {n_bursts_axis_None}
+    Number of bursts (axis = 0   ): {n_bursts_axis_0}
+""".format(n_bursts_axis_None=len(df_bursts), n_bursts_axis_0=len(df_bursts_axis0)))
+
+####################################################################################################
+#
+# Example 2. 2D Array (n_copies, n_timepoints)
+# --------------------------------------------
+# Features for a 2D array with an (n_copies, n_timepoints) organization will be computed
```
TomDonoghue, line 133: You're section-opening sentences read as slightly hard to parse to me. Like, they feel almost like they are missing works (shorthand), though I think what it actually is is that the verb (compute) & object (features) are super far apart, which can make it slightly hard to track, unless one already knows what's being said. 

Like, here I'd write it something like:
"In this example, we will _compute_ features for a 2D array in which we have organized multiple copies of the same signal, to explore applying different thresholds"

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
...
+sim_components = {'sim_powerlaw': dict(exponent=-2),
+                  'sim_bursty_oscillation': dict(freq=10)}
+
+sig = sim_combined(n_seconds, fs, components=sim_components)
+
+sigs = np.array([sig] * n_copies)
+
+####################################################################################################
+
+# Step amp_consistency thresholds by 0.25
```
TomDonoghue, line 150: step _through_ (?)

examples/plot_2_multidim_sigs.py
```diff
@@ -0,0 +1,249 @@
...
+thresholds = dict(amp_fraction_threshold=0., amp_consistency_threshold=.5,
+                  period_consistency_threshold=.5, monotonicity_threshold=.8,
+                  min_n_cycles=2)
+
+amp_consist_threshes = [0, .25, .5, .75, 1.]
+
+compute_kwargs = [{'threshold_kwargs': dict(thresholds.copy(), amp_consistency_threshold=thresh)}
```
TomDonoghue, line 157: doesn't copy return a dict? Why do you need to typecast?

TomDonoghue (changes requested):

TomDonoghue:
I did a review through, and made a bunch of suggestions to the main new files here. 

A broader thought - I actually think the new example for multi-dim could be two different examples? I don't think there's any strong reason 2D and 3D have to be grouped. I think two shorter and more specific examples might work a bit better. Also: maybe these are _tutorials_, rather than _examples_? I'm not sure, but I could see these being built into the tutorial layout, as numbers 4 and 5, rather than as standalone examples. What do you think @ryanhammonds ?

doc/changelog.rst
ryanhammonds: All related changes to chaneglog.rst in in 8a4c036.

examples/plot_2_multidim_sigs.py
ryanhammonds: Update in 8619a51.

ryanhammonds:
Thanks for the review, @TomDonoghue. I made changes based on all of you feedback. I'm going to merge this in but will go back through all of the 1.0.0 changes together for typos/cleans/etc.

TomDonoghue:
@ryanhammonds - sounds good! I had a super quick scroll here, and it all looks good to me! 
And yeh, we can do a final sweep of all things for 1.0 after it's all settled.

Issue #81, opened 2020-11-28
reduce size of example data files


TomDonoghue:
The data files in `examples/data/` are pretty big files (~47 MB each). These big files make the repo needlessly big to clone / download, etc. 

From a quick check, we only use these files once, in the `plot_theta_feature_distributions` example. Within that example, we immediately extract 1 second of data (about 1% of the length), so most of this data isn't even used in the example. 

Basically, we should trim and replace the data files and only package the 1 second segments in `examples/data/`. 

@ryanhammonds - can you double check if these files are used anywhere else? If not, please make the minimized data files, and replace in `examples`. 

Sidenote: I wouldn't be opposed to keeping a collection of some bigger datafiles that are used for examples / testing, etc - but if so, these could be put in a separate place / repo (`bycycle-tools/Data`, for example), to still keep the code repo as small.

ryanhammonds:
Complete in #82.

PR #82, opened 2020-12-10
[ENH] Reduce Example Filesize

ryanhammonds:
Address #81. This PR reduces ` ca1.npy` and `ec3.npy` to 588kb (from 48mb). Also, `plot_2_bycycle_algorithm.py ` use to rely on  `ca1.npy` before it was replaced with simulated data. I reverted this back since simulating the data sometimes produced funky cyclepoints.

I also add util functions to download example data (same as in ndsp).


tutorials/plot_2_bycycle_algorithm.py
```diff
@@ -51,18 +51,15 @@
...
+from bycycle.utils.download import load_bycycle_data
 
 pd.options.display.max_columns = 10
 
 ####################################################################################################
 
-# Simulation settings
-n_seconds = 100
+# Load data
+sig = load_bycycle_data('ca1.npy', folder='../examples/data')
```
TomDonoghue, line 61: i think this folder can / should be just 'data'? Like it is in the `plot_1_theta` tutorial? Otherwise, doesn't this, from inside an 'examples' folder, create a en embedded folder of `examples/examples/data/`?

TomDonoghue (approved):

tutorials/plot_2_bycycle_algorithm.py
ryanhammonds: If `data` is used in this case, we'll end up with the same data in two locations (`tutorials/data` and `examples/data`).  The path grabs the data from examples for tutorials.

TomDonoghue: Oh, I didn't notice the '..' - but also, i don't think "same data in two locations" actually matters - because that's just temporary. The goal of doing this is if you download this file (independently) it can download and load the data you need. Right now, this would do so with a weird (or possible breaking) path location. 

Which actually makes me notice something: the data files should be moved to a base folder 'data' in the bycycle repo, just like in neurodsp. 

In the end, this does mean that building the whole site would lead to duplicated data files - but that's just for a temporary build. The data actually live in one place, (`bycycle/data`), and can be accessed from any tutorial / example that needs it, and I think should use a sensible / local data folder when it does so. 

Does that make sense?

ryanhammonds: I see, I didn't think about people downloading this file independently. I'll push updates for this.

ryanhammonds: I update `../examples/data` to `data` and moved the data folder to `bycycle/data` in the latest commit.

TomDonoghue:
It's normal this fails after moving data files, as the files are not in the expected place on master. It will pass once merged. 

This all looks good to me!

PR #88, opened 2021-02-03
[BUG] Limit Dataframe Fix

ryanhammonds:
There was a bug when using `xlim` with plotting funcs. Is was a mix up of `df['sample_last_' + side_e]` and `df['sample_next_' + side_e]` in the `limit_df` func. Here's an example:

```python
import numpy as np
from bycycle.features import compute_features
from bycycle.plts import plot_burst_detect_summary

n_seconds = 10
fs = 1000

sig = np.random.rand(int(n_seconds * fs))
times = np.arange(0, len(sig)/fs, 1/fs)

thresholds = {'amp_fraction_threshold': 0.5,
              'amp_consistency_threshold': 0.5,
              'period_consistency_threshold': 0.5,
              'monotonicity_threshold': 0.75,
              'min_n_cycles': 3}

df_features = compute_features(sig, fs, (1, 50), threshold_kwargs=thresholds)
plot_burst_detect_summary(df_features, sig, fs, thresholds, xlim=(1, 2))
```

PR #90, opened 2021-02-04
[ENH] Recompute consistency features at burst edges

ryanhammonds:
This is an improvement to #73 and `recompute_edges`. Period/amplitude consistency features may now be computed in one direction for non-bursting cycles that are immediately adjacent to bursts. Before the solution was to drop thresholds for edge cycles which works but was lazier.

```python
import numpy as np
from neurodsp.sim import sim_combined

from bycycle.features import compute_features
from bycycle.burst.utils import recompute_edges
from bycycle.plts import plot_burst_detect_summary

# Simulate
np.random.seed(0)
n_seconds = 10
fs = 1000
f_range = (1, 51)

components = dict(sim_bursty_oscillation=dict(freq=15, cycle='gaussian', std=1),
                  sim_powerlaw=dict(exponent=-2))

sig = sim_combined(n_seconds, fs, components)

times = np.arange(0, len(sig)/fs, 1/fs)

thresholds = {'amp_fraction_threshold': 0.3,
              'amp_consistency_threshold': 0.55,
              'period_consistency_threshold': 0.55,
              'monotonicity_threshold': 0.75,
              'min_n_cycles': 3}

df_features = compute_features(sig, fs, f_range, threshold_kwargs=thresholds)

plot_burst_detect_summary(df_features, sig, fs, thresholds, plot_only_result=True, xlim=(5.8, 6.3))

indices = np.where((df_features['sample_peak'] >= (5.9* fs)) &
                   (df_features['sample_peak'] <= (6.2* fs)))[0]

print('Original:   ', df_features.iloc[indices]['period_consistency'].values)

df_features = recompute_edges(df_features, thresholds)

plot_burst_detect_summary(df_features, sig, fs, thresholds, plot_only_result=True, xlim=(5.8, 6.3))

print('Recomputed: ', df_features.iloc[indices]['period_consistency'].values)
```
consist

### DIFF
--- a/bycycle/burst/utils.py
+++ b/bycycle/burst/utils.py
@@ -2,7 +2,7 @@
 
 from copy import deepcopy
 import numpy as np
-from bycycle.utils.checks import check_param
+from bycycle.utils.checks import check_param_range, check_param_options
 
 ####################################################################################################
 ####################################################################################################
@@ -32,7 +32,7 @@ def check_min_burst_cycles(is_burst, min_n_cycles=3):
     """
 
     # Ensure argument is within valid range
-    check_param(min_n_cycles, 'min_n_cycles', (0, np.inf))
+    check_param_range(min_n_cycles, 'min_n_cycles', (0, np.inf))
 
     temp_cycle_count = 0
 
@@ -136,6 +136,9 @@ def recompute_edge(df_features, cyc_idx, direction):
     df_features : pandas.DataFrame
         A dataframe with updated consistency features.
     """
+
+    # Check that param validity
+    check_param_options(direction, 'direction', ['both', 'next', 'last'])
 
     # Prevent circular imports between burst.utils and burst.cycle
     from bycycle.features.burst import compute_amp_consistency, compute_period_consistency
--- a/bycycle/burst/utils.py
+++ b/bycycle/burst/utils.py
@@ -94,7 +94,6 @@ def recompute_edges(df_features, threshold_kwargs, burst_method='cycles', burst_
 
     # Prevent circular imports between burst.utils and burst.cycle
     from bycycle.burst import detect_bursts_cycles
-    from bycycle.features.burst import compute_amp_consistency, compute_period_consistency
 
     # Prevent overwriting the original dataframe
     df_features_edges = df_features.copy()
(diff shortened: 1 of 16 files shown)


bycycle/burst/utils.py
```diff
@@ -108,6 +109,28 @@ def recompute_edges(df_features, threshold_kwargs):
...
     # Recompute is_burst
+    for idx, cyc_idx in enumerate(burst_edges):
+
+        if idx % 2 == 0:
+            direction = 'next'
+        else:
+            direction = 'last'
+
+        # Slice edges rows and recompute burst features
+        lower = cyc_idx-1 if cyc_idx-1 >= 0 else 0
```
srcole, line 120: nit: `max(cyc_idx-1, 0)` is slightly simpler

bycycle/burst/utils.py
```diff
@@ -108,6 +109,28 @@ def recompute_edges(df_features, threshold_kwargs):
...
+        upper = cyc_idx+2 if cyc_idx+2 < len(df_features_edges) else len(df_features_edges)
```
srcole, line 121: similar idea as last comment. something like `min(cyc_idx+2, len(df_features_edges))`

bycycle/features/burst.py
```diff
@@ -127,15 +128,15 @@ def compute_amp_fraction(df_shape_features):
...
+def compute_amp_consistency(df_shape_features, direction=None):
     """Compute amplitude consistency for each cycle.
 
     Parameters
     ----------
     df_shape_features : pandas.DataFrame
         Shape features for each cycle, determined using :func:`~.compute_shape_features`.
-    df_samples : pandas.DataFrame
-        Indices of cyclepoints returned from :func:`~.compute_cyclepoints`.
+    direction : {'next', 'last'}
```
srcole, line 138: nit: I'd prefer stylistically if the default was the same datatype as the defined options, i.e. {'next', 'last', 'both'}.

srcole: I think there should also be a check that one of these values was inputted (and so, for example, an error is raised if someone misspells `next` as `nxet`, which would currently be treated as if it was `None`

bycycle/burst/utils.py
```diff
@@ -108,6 +109,28 @@ def recompute_edges(df_features, threshold_kwargs):
                             enumerate(burst_edges)])
 
     # Recompute is_burst
+    for idx, cyc_idx in enumerate(burst_edges):
+
+        if idx % 2 == 0:
```
srcole, line 114: this is a little opaque to me. Why do we do "next" if this index is even? Is it because that would indicate the start of a burst? If so, I think the code could be written in a way that this is more obvious.

Like instead of `burst_edges` containing both the starting and stopping edges of a burst, maybe it'd be better to refactor this to be 2 lists of the same length (e.g. burst_starts and burst_ends). That's just one option. Whatever you think is best.

bycycle/utils/dataframes.py
```diff
@@ -55,10 +55,10 @@ def limit_df(df, fs, start=None, stop=None):
 
     start = 0 if start is None else start
 
-    df = df[df['sample_next_' + side_e].values >= start*fs]
+    df = df[df['sample_last_' + side_e].values >= start*fs]
```
srcole, line 58: I'm guessing these changes should be removed because they're in #88 ?

bycycle/burst/utils.py
ryanhammonds: Thanks for taking the time to review this!

Yeah, `burst_edges` alternates between cycles before/after bursts. I like your idea of having 2 lists - that would make things clearer. I'll push that change soon.

srcole:
This is a really good change! Nice example in the first comment too.

bycycle/burst/utils.py
ryanhammonds: That's a much simpler than what I have, thanks!

bycycle/features/burst.py
ryanhammonds: I like this idea. I'll set this to default to 'both'.

I wrote an updated check func in neurodsp/utils/checks.py (unmerged in open PR) that does exactly your second comment. That could be imported here. I need to do a sweep of the whole code base, because there are probably other places type checking could be helpful.

bycycle/utils/dataframes.py
ryanhammonds: Good catch! The first two commits slipped in here by accident. I'll drop them.

ryanhammonds:
Thanks for the review! I pushed related updates. The last thing to do is sort out type checking.

srcole (approved):
looks good!

Issue #99, opened 2021-03-19
v1.1.0


ryanhammonds:
Minor release tentative plans:

- [x] Bycycle and BycycleGroup classes (#98)
- [x] ~~Spike extension (#97, #100)~~
- [ ] Global burst threshold (#89)
- [x] Doc updates (#84, #105)
- [x] Bugs (#92, #94)

ryanhammonds:
@voytek I can use this and the 1.0.0 changelog for the progress report.

Issue #97, opened 2021-03-12
Features for single cycles


ryanhammonds:
Running bycycle on single cycles, from spike sorting or erp data, has been hacky since the signal has to be padded with zeros so filtering won't raise an error and some features (i.e. period/amp consistency) don't make sense in these contexts. I have some code that could be adapted into a sub-module for this purpose.

PR #100, opened 2021-04-08
[MRG] Spikes

ryanhammonds:
Addresses #97 (and #99).

```python
from neurodsp.filt import filter_signal
from bycycle import Spikes

# The spiking data is from: https://zenodo.org/record/1205233#.YHYezhJlCWh
# The links includes instructions on how to load data into np arrays. 

f_ranges = (300, 2000)
sig_filt = filter_signal(sig, fs, 'bandpass', f_range, remove_edges=False)
sig_filt = sig_filt[:int(fs* 5)]

spikes = Spikes()
spikes.fit(sig_filt, fs, f_range)

# Plot 2d stack
spikes.plot()

# Plot single spike
spikes.plot(index=0)

# Plot full signal with spikes highlighted
spikes.plot(stack=False, xlim=(0, 1))

# Access dataframe of shape features
spikes.df_features
```
new_spikes

To-Do:
- [ ] Tests


ryanhammonds:
Another to-do: add a class/func to use bycycle to segment spiking data into 2d arrays, where each 1d array is a putative spike. I have a notebook where I've done this and was able to cluster/sort spikes. This can be added as a tutorial too.   
spike_clusters

PR #115, opened 2021-10-08
[MNT] - Add doctest testing to Github Actions

TomDonoghue:
Add testing of doctest examples in Github actions (same as https://github.com/neurodsp-tools/neurodsp/pull/278).


TomDonoghue:
This is failing because of unexpected printed output from running a test without having `tqdm` installed. 

Though we could tweak this doctest to pass, I think it more generally suggests we should run tests with optional requirements installed, which is PR'd in #116, after which I think this PR should also pass.

Issue #128, opened 2023-01-18
Documenting code fixes


dillancellier:
- [ ] *features/burst.py/compute_monotonicity* : might be able to change `monotonicity = np.ones(cycles) * np.nan` so that the np.ones is not being initialized just to be replaced by nans
- [ ] *features/burst.py/compute_monotonicity* : can convert dataframe to dictionary temporarily to loop through rows. df.iterrows() seems to be slowest possible option (which is what the code currently does) see this page-- https://towardsdatascience.com/heres-the-most-efficient-way-to-iterate-through-your-pandas-dataframe-4dad88ac92ee

mwprestonjr:
Other functions in features/burst.py have the the same initialization issue as compute_monotonicity() (line 288), including compute_amp_consistency() (line 163) and compute_period_consistency() (line 239),

dillancellier:
- [ ]  *features/burst.py/compute_amp_consistency* : might be able to change all instances of np.nanmin or nanmax to commands  using Bottleneck https://pypi.org/project/Bottleneck/0.4.2/ but might be an issue of adding dependencies?

PR #129, opened 2023-01-18
[MRG] Optimize Runtime

ryanhammonds:
Optimizations to decrease runtime. Related to #128.

### DIFF
--- a/bycycle/cyclepoints/zerox.py
+++ b/bycycle/cyclepoints/zerox.py
@@ -69,7 +69,7 @@ def find_zerox(sig, peaks, troughs):
     return rises, decays
 
 
-def find_flank_zerox(sig, flank):
+def find_flank_zerox(sig, flank, midpoint=None):
     """Find zero-crossings on rising or decaying flanks of a filtered signal.
 
     Parameters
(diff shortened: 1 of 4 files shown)


ryanhammonds:
One thing I've been thinking of for this PR, and other packages in general is how to ensure consistency of outputs for each PR merge, relative to the last pypi release. Storing arrays of results and comparing seems cumbersome on git tracking. An option may be to track the hash of a np array (commit it per release) and compare it in a unit test for PRs.

ryanhammonds:
This is ready to merge. Tests are passing, and the the new consistency test in #131 passes - meaning the dataframes results (based on the simulated signal in the pytest fixture) are exactly equal between from this PR to the past version.

I ran a few benchmarks using this version vs the last pypi version, with a simulated powerlaw + 10hz oscillation. This PR sped up `compute_features` by up to 32% faster but only for long signals (10 million samples). Shorter signals (10k samples) didn't see improvements:

#### Signal with 10m Samples

1 signal (compute_features):

> Before: 11.4 s ± 14.2 ms
> After : 7.8 s ± 30.2 ms
> ~32% faster

6 signals in parallel (compute_features_2d):

> 16.1 s ± 590 ms
> 11.7 s ± 120 ms
> ~27% faster

#### Signal with 10k samples

1 signal (compute_features):

> Before: 13.7 ms ± 137 µs
> After : 13.8 ms ± 99.8 µs

6 signals in parallel (compute_features_2d):

> Before: 1.39 s ± 62.8 ms
> After : 1.34 s ± 49.2 ms

Issue #135, opened 2023-06-10
Conda install version `1.0.0` not compatible with `BycycleGroup` tutorials


ntolley:
Just came across the issue that some of the tutorials are not compatible `1.0.0` release distributed on the conda repository. Specifically it looks like the `BycycleGroup` class wasn't added until `1.1.0` so it throws an import error. pip install works fine of course!

ryanhammonds:
I merged this [PR](https://github.com/conda-forge/bycycle-feedstock/pull/2) that may have fixed the issue? I don't use conda but lmk if it doesn't fix the install.

ntolley:
Yeah that updated the version! Can confirm the import passes after testing locally.

Also related to the same tutorial, it seems like there's a variable that doesn't get used [here](https://github.com/bycycle-tools/bycycle/blob/44c995d06fcd2e746fd4e1955d35f4a69c64fbaf/tutorials/plot_5_3D_sigs.py#L116)

I can try revising it in a PR if you guys are are cool with it! I'm guessing it should be something like:
```py
compute_kwargs = {'burst_method': 'cycles', 'thresholds': thresholds}
bg_rest = BycycleGroup(**compute_kwargs)
```
?

ryanhammonds:
@ntolley I can review a PR for this is you want to make one. Thanks!

PR #137, opened 2023-06-13
[MRG] Doc: clean up tutorial code in `ByGroup` tutorial

ntolley:
Discussed in #135 

@ryanhammonds let me know if this looks good with you! Alternative is to just be explicit and provide the parameters as named arguments so I'll let you decide.


ryanhammonds:
Thanks for making this PR! This looks good to me, unpacking the dict makes sense since those params are reused. Merging this in

PR #143, opened 2023-10-29
[MNT] - Update workflow tests to support Python 3.12

TomDonoghue:
This PR updates the workflow to test against 3.12, which is now available. 

Confusingly, in the workflow tests, version below 3.12 (which previously passed) are now failing - which is not happening when I run tests locally. It looks like (at least part of this) might be a plotting thing, as one of the error messages is about markerstyle. @ryanhammonds - any ideas what is happening here?

Edit - this was previously failing, but it turns out it was about #144 - now that that is merged here, it works!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
       MODULE_NAME: bycycle
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


ryanhammonds (approved):
Sorry I missed this! Thanks for updating support for 3.12. Glad tests are passing now.

PR #144, opened 2023-12-09
[BUG] - Update marker definition for newer matplotlib

TomDonoghue:
Issue: for current matplotlib version >= 3.8, they have deprecated passing in a marker definition as None to line style (see note in docs [here](https://matplotlib.org/stable/api/markers_api.html)). 

This update fixes this, so that None -> "None", which is interpreted as no marker style. 

Note / to check: I'm not sure if "None" will work on all older version of matplotlib - it is possible this changes behavior / creates an error on older matplotlibs.

### DIFF
--- a/bycycle/plts/burst.py
+++ b/bycycle/plts/burst.py
@@ -242,7 +242,7 @@ def plot_burst_detect_param(df_features, sig, fs, burst_param, thresh,
 
         plot_time_series([times[df_features['sample_' + center_e]], (times[0], times[-1])],
                          [df_features[burst_param], [thresh]*2], ax=ax, colors=['k', 'k'],
-                         ls=['-', '--'], marker=["o", None], xlabel=xlabel,
+                         ls=['-', '--'], marker=["o", "None"], xlabel=xlabel,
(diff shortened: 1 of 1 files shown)


ryanhammonds:
> I'm not sure if "None" will work on all older version of matplotlib - it is possible this changes behavior / creates an error on older matplotlibs.

Should we pin matplotlib to >= latest in the requirements to prevent older versions from raising errors with "None"?

ryanhammonds (approved):

TomDonoghue:
@ryanhammonds - I don't actually know if older matplotlib's will have an issue (I have no idea when "None" was accepted - maybe from the beginning). To get a quick sense - I ran the tests against matplotlib 3.1.0, the start of the current major version of mpl (for some reason 3.0 wouldn't install) and a version going back about 4 years, and it all worked as is. With that, I think I'd say we don't need to lock to newest matplotlib, nor try and figure out any specific minimum version - I think we can probably leave as is?

davidglitvin:
I rolled back to matplotlib 3.7.1 and everything worked fine.

PR #151, opened 2025-01-28
debug marker style error

mzabolocki:
Fixed an error with the parsed marker style by removing 'none' from the marker parameters. The plot_time_series function now works as expected, especially for the interp option.


ryanhammonds:
Thanks for the fix!

